PR #3, opened 2016-03-09
Start extracting functionality

derwolfe:
This is the first pass at pulling out logic.

Includes
- get-auth-token! - this is meant to fetch a token from their api
  using the secret key and id provided by cp. How this is persisted
  is up to the consumer.
- get-single-events-page! - used to issue requests against the cp
  api.
- ->cp-date - function for formatting dates in way that their api
  expects.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ cache:
 script:
 - lein cljfmt check
 - lein test
-- lein uberjar
 after_success:
 - CLOVERAGE_VERSION=1.0.7-SNAPSHOT lein cloverage --codecov
 - bash <(curl -s https://codecov.io/bash) -f target/coverage/codecov.json
--- a/project.clj
+++ b/project.clj
@@ -1,16 +1,21 @@
 (defproject cloudpassage-lib "0.1.0-SNAPSHOT"
-  :description "Clojure interface for the CloudPassage API"
-  :url "https://github.com/RackSec/cloudpassage-lib"
+  :description "A library for interacting with cloudpassage apis."
+  :url "http://github.com/RackSec/cloudpassage-lib"
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :dependencies [[org.clojure/clojure "1.8.0"]]
-  :plugins [[lein-cljfmt "0.3.0"]
+  :dependencies [[org.clojure/clojure "1.8.0"]
+                 [com.taoensso/timbre "4.2.0"]
+                 [manifold "0.1.2"]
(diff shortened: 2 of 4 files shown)


project.clj
```diff
@@ -1,16 +1,20 @@
 (defproject cloudpassage-lib "0.1.0-SNAPSHOT"
-  :description "Clojure interface for the CloudPassage API"
-  :url "https://github.com/RackSec/cloudpassage-lib"
-  :license {:name "Eclipse Public License"
-            :url "http://www.eclipse.org/legal/epl-v10.html"}
-  :dependencies [[org.clojure/clojure "1.8.0"]]
-  :plugins [[lein-cljfmt "0.3.0"]
+  :description "A library for interacting with cloudpassage apis."
+  :url "http://github.com/RackSec/cloudpassage-lib"
+  :license {:name "proprietary"}
```
ehashman: I think you accidentally changed this line.

derwolfe: Indeed. I'll remove this :-).

.travis.yml
```diff
@@ -8,7 +8,6 @@ cache:
 script:
 - lein cljfmt check
 - lein test
-- lein uberjar
```
ehashman: We're getting rid of the jar step because this is a library and not a single executable, is that right?

derwolfe: Yes.

src/cloudpassage_lib/core.clj
```diff
@@ -1 +1,73 @@
-(ns cloudpassage-lib.core)
+(ns cloudpassage-lib.core
+  (:require
+   [clojure.string :as str]
+   [aleph.http :as http]
+   [manifold.deferred :as md]
+   [clj-time.format :as f]
```
ehashman: Do you want to put this below clj-time.core?

src/cloudpassage_lib/core.clj
```diff
@@ -1 +1,73 @@
...
+   [manifold.time :as mt]
+   [byte-streams :as bs]
+   [taoensso.timbre :as timbre :refer [info error]]
+   [base64-clj.core :as base64]
+   [clj-time.core :as time]
+   [cheshire.core :as json]))
+
+;; the url from which new auth-tokens can be obtained.
+(def auth-uri "https://api.cloudpassage.com/oauth/access_token?grant_type=client_credentials")
```
ehashman: This was `defonce` before -- could you tell me about the difference between that and `def`?

derwolfe: I changed these to defs so that consumer of the library could change them as they see fit. If they are defined using defonce, then the bindings cannot be re-bound to contain different values.

PR #4, opened 2016-03-09
add redis and fernet

derwolfe:
This adds support for persisting auth tokens, encypted using fernet, to redis.


ehashman:
Hm, wonder why we don't have gating coverage. LGTM.

PR #6, opened 2016-03-09
Add some docs

ehashman:


README.md
```diff
@@ -1,14 +1,69 @@
...
+   ```
+
+   to fetch dependencies.
+
+3. Obtain a fernet key. You only need to generate this once; you can store it
+   in your `.lein-env` later, per step 2.
+
+   ```clojure
+   lein repl
+   cloudpassage-poller.core=> (require 'fernet.core)
```
derwolfe: s/poller/lib

README.md
```diff
@@ -1,14 +1,69 @@
-# cloudpassage-lib
+# cloudpassage-lib #
```
derwolfe: Should there be a trailing #?

doc/cloudpassage-api.md
```diff
@@ -0,0 +1,71 @@
...
+
+3. When making a call to the API, pass the access token in the Authorization
+   header field. Specify the authentication scheme as "Bearer", followed by a
+   space and then the token. For example:
+
+   ```
+   Authorization: Bearer ffad76cc550110fc4c84a18397b6e104
+   ```
+
+## Events ##
```
derwolfe: Since we aren't exposing anything specific to events, I think this could be removed.

README.md
ehashman: It's a style thing, I prefer it that way because it's symmetric and I'm silly :) The trailing # doesn't render.

ehashman: Made the other fixes but didn't see this early enough to include it. Do you want me to go back and make this change as well?

derwolfe: Yes, please. :-)

PR #11, opened 2016-03-11
Implement scans-with-details! and consume in fim-report!

### LINKED ISSUE #10: fim-report! should fetch more data
Currently the output of `fim-report!` looks like this:

```clojure
({:server_id "09d36abea5cc11e591527d9f85f6c9bc",
  :server_url
  "https://api.cloudpassage.com/v1/servers/09d36abea5cc11e591527d9f85f6c9bc",
  :completed_at "2016-03-07T18:45:59.840Z",
  :server_hostname "111111-cf01",
  :module "fim",
  :non_critical_findings_count 0,
  :status "completed_clean",
  :id "d4d0caace49411e58d021b460156fb0c",
  :ok_findings_count 3801,
  :url
  "https://api.cloudpassage.com/v1/scans/d4d0caace49411e58d021b460156fb0c",
  :critical_findings_count 4,
  :created_at "2016-03-07T18:45:58.006Z"}
...)
```

This probably isn't detailed enough for compliance purposes. It looks like the rest of the scan data lives at the `:url`; we should probably improve `fim-report!` to fetch and return the data located there as well.

ehashman:
Fixes #10.

I think that we need to go one level deeper on the scan, so the next step might be to implement `scans-with-even-more-details!` or the like.

### DIFF
--- a/src/cloudpassage_lib/scans.clj
+++ b/src/cloudpassage_lib/scans.clj
@@ -14,7 +14,7 @@
    [clj-time.format :as tf]
    [clojure.java.io :as io]))
 
-(def base-scans-url
+(def ^:private base-scans-url
   "https://api.cloudpassage.com/v1/scans/")
 
 (def ^:private scans-path
--- a/src/cloudpassage_lib/scans.clj
+++ b/src/cloudpassage_lib/scans.clj
@@ -91,7 +91,14 @@
   more illustration."
   [client-id client-secret scans-stream]
   (let [scans-with-details (ms/stream 10)
-        add-details (fn [scan])] ;; TODO: actually add details ;)
+        add-details
+        (fn [scan]
+          (md/chain
+           (get-page! client-id client-secret (:url scan))
+           (fn [response]
+             (ms/put! scans-with-details (assoc scan :scan (:scan response)))
+             (when (ms/drained? scans-stream)
+               (ms/close! scans-with-details)))))]
     (ms/consume-async add-details scans-stream)
     scans-with-details))
 
@@ -100,5 +107,6 @@
   [client-id client-secret]
   (let [opts {"since" (cpc/->cp-date (-> 3 hours ago))}]
     (->> (scans! client-id client-secret opts)
-         ms/stream->seq
-         (filter (fn [{:keys [module]}] (= module "fim"))))))
+         (ms/filter (fn [{:keys [module]}] (= module "fim")))
+         (scans-with-details! client-id client-secret)
+         ms/stream->seq)))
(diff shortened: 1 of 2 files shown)


ehashman:
(Reminder to myself: once this gets merged, we should cut a new minor release.)

src/cloudpassage_lib/scans.clj
```diff
@@ -14,7 +14,7 @@
    [clj-time.format :as tf]
    [clojure.java.io :as io]))
 
-(def ^:private base-scans-url
+(def base-scans-url
```
derwolfe: Just out of curiousity - why is this now public api?

test/cloudpassage_lib/scans_test.clj
```diff
@@ -69,12 +78,18 @@
       (let [since (tf/parse cp-date-formatter since)
             fudged (t/interval (-> 4 hours ago) (-> 3 hours ago))]
         (is (within? fudged since))))
-    {:scans (for [index-in-page (range scans-per-page)
-                  :let [scan-index (+ (* scans-per-page page-num)
-                                      index-in-page)]]
-              {:scan-id scan-index
-               :module (index->module scan-index)})
-     :pagination {:next next-page}}))
+    (if (query "details")
```
derwolfe: Does the `scans-with-details` code need any of the pagination specific code in lines 71 though 80?

src/cloudpassage_lib/scans.clj
ehashman: I used it in the tests. It used to be public, then I think it became private in another PR, and now I've made it public again. Not sure if there's a better way.

derwolfe: If you'd like to keep it private, but need to use it your tests, you can varquote it via `#'scans/base-scans-url`. Some examples of this are available inside of `scans-test`, (https://github.com/RackSec/cloudpassage-lib/blob/master/test/cloudpassage_lib/scans_test.clj#L14)

test/cloudpassage_lib/scans_test.clj
derwolfe: Whoops - that comment was cutoff for some reason. Here is the rest: 

I ask because I'm wondering this should just be a separate function :-).

ehashman: `scans/scans-with-details!` consumes a stream whose pagination has already been unrolled by `scans!`. I'm not sure if that answers your question, though.

I'm a bit grumpy how git displays this diff. The "removed" code above is only reading that way because of the whitespace diff; the insertion of the `if (query "details")` and first case below (the details scans are not paginated, so I wanted to reflect this in `fake-get-page!`) is the only difference.

derwolfe:
Thanks! This is ready to land.

Issue #12, opened 2016-03-11
Refactor the library such that it doesn't need to look at environment variables


ehashman:
Here's an example of an error message that doesn't say very much:

```
=> (scans/fim-report! "$ID" "$KEY")
16-03-11 20:25:17 MJV0HLDKQ4 INFO [cloudpassage-lib.core] - fetching new auth token for $ID
Mar 11, 2016 2:25:17 PM clojure.tools.logging$eval420$fn__424 invoke
SEVERE: error in stream handler
java.lang.NullPointerException
    at java.util.Arrays.copyOfRange(Arrays.java:3521)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:497)
    at clojure.lang.Reflector.invokeMatchingMethod(Reflector.java:93)
    at clojure.lang.Reflector.invokeStaticMethod(Reflector.java:207)
    at fernet.core$split_key.invokeStatic(core.clj:29)
    at fernet.core$split_key.invoke(core.clj:28)
    ...
```

Yay NPE!

The actual problem here was that the fernet key/redis environment variables weren't set properly in profiles.clj. It would be great to do some checking for that and provide a more user-friendly error message.

derwolfe:
Thanks for reporting this issue. :-)

I'm a bit unsure that this is a library level concern; in fact, I'm inclined to say that the library shouldn't be consuming environment variables at all. Consumers of the library should be passing parameters to functions.

The reason the current code is relying on environment variables is because I extracted the logic directly from an application without alteration.

Currently, the library relies on the following variables to function
1. Redis-url (basically a db connection string)
2. Redis-timeout
3. Fernet-key

`scans/get-page!` does indeed rely on `fernet-key` existing in the environment; in my opinion, this should removed in favor of passing a parameter into the function. The reason I prefer this approach is because it pushes the input validation onto the application. Any consuming application will have a set of environment variables that should be set in order to function; further, this allows the application to determine what sort of remediation should occur in the event of a bad env-var having been used.

[...]

PR #15, opened 2016-03-11
Better error messages WRT undefined env values

TimSimpsonR:
Adds some code to check if a value from env is nil or blank before
returning it.

Fixes issue #12

### DIFF
--- a/src/cloudpassage_lib/core.clj
+++ b/src/cloudpassage_lib/core.clj
@@ -2,7 +2,7 @@
   (:require
    [clojure.string :as str]
    [aleph.http :as http]
-   [environ.core :refer [env]]
+   [environ.core :as env-core]
    [manifold.deferred :as md]
    [manifold.time :as mt]
    [byte-streams :as bs]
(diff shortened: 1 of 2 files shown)


src/cloudpassage_lib/core.clj
```diff
@@ -14,13 +14,36 @@
...
+
+(deftype CheckEnv
+  []
+  clojure.lang.ILookup
+  (valAt [this k not-found]
+    (env-get k))
+  (valAt [this k]
+    (.valAt this k nil)))
+
+(def conf (CheckEnv.))
```
derwolfe: @TimSimpsonR - It seems like there should a be a simpler solution that just uses functions to do this, instead of needing to add a new type and some more associated state. Would you mind explaining why this is necessary?

TimSimpsonR: Maybe there is, but I couldn't make a function allow for de-structuring, i.e 
```clojure
 (let [{:keys [redis-url redis-timeout]} conf]
...
```
Wouldn't work with just a function, as when a key is used as a function and passed another function it simply returns nil.

derwolfe: (discussed in slack)

There are a few ways this could done. The prime factor seems to be how often/when you want to perform validation. If this is a startup step only; there could just be a function that is invoked that asserts/checks that certain values are not nil or conform to some other requirement. If they don't, explode and no longer work; if the values are valid, then execute as if nothing happened. Another option would be to validate on each access to the env vars. 

I'm solidly in the camp for the former :-).

ehashman: +1 to a startup step that ensures certain values aren't nil!

derwolfe:
A few points arose during our discussion:
1) Should the library be validating specific keys? Or should there be a more general environment validation function that takes a list of keys and blows up when there are errors. This is a library and might not contain all of the environment variables that consumer applications are using.
2) Do we want to create a new piece of state to contain configuration values?
3) Should we redesign how the application is accessing redis? It seems like there might be reasons to pass parameters around, instead of relying on environment state.

TimSimpsonR:
I am totally down with #3. I really want to make this trip to Clojure town. I kind of feel like technicalities aside the env stuff is a global var and we should reduce or eliminate its use anywhere outside of non-library code (and maybe even then we could stick to main methods only).

TimSimpsonR:
I'm closing this as I don't believe in it.

PR #18, opened 2016-03-14
Adds SVM reporting capabilities

TimSimpsonR:


src/cloudpassage_lib/scans.clj
```diff
@@ -102,11 +102,21 @@
...
 
-(defn fim-report!
-  "Get the current (recent) FIM report for a particular client."
-  [client-id client-secret]
+(defn report-for-module!
+  "Get recent report data for a certain client, and ilter based on module."
+  [client-id client-secret module-name]
   (let [opts {"since" (cpc/->cp-date (-> 3 hours ago))}]
     (->> (scans! client-id client-secret opts)
-         (ms/filter (fn [{:keys [module]}] (= module "fim")))
```
ehashman: So this filtering is specifically required for FIM because the CloudPassage API is broken. You *should* be able to just call `scans!` with the correct query URL for SVA/CSM and have it "Just Work."

TimSimpsonR:
This is ready to review. @ehashman and @derwolfe if you could look at this when you have the time.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -115,29 +115,44 @@
...
-          scans (ms/stream->seq scans-with-details)]
-      (is (= (for [scan-id (range (* fake-pages scans-per-page))]
-               {:scan-id scan-id
-                :module (index->module scan-id)
-                :url details-query-url
-                :scan {}})
-             scans))
-      (is (ms/closed? scans-stream))
-      (is (ms/closed? scans-with-details)))))
+  (testing "Typical scan"
```
derwolfe: What would you think about changing `Typical scan` to more of an assertion - like `Typical scan returns ...`

test/cloudpassage_lib/scans_test.clj
```diff
@@ -115,29 +115,44 @@
...
+  (testing "Blank input stream won't block."
+    (let [empty-stream (ms/stream 0)]
+      (ms/close! empty-stream)
+      (->> empty-stream
+           (scans/scans-with-details! "client-id" "client-secret")
+           (ms/stream->seq)
+           doall))))
 
-(deftest fim-report!-tests
+(defn test-report
```
derwolfe: Would you mind marking this as `^:private`? It's purely a convention, but many of the other test helper functions are marked as such.

src/cloudpassage_lib/scans.clj
```diff
@@ -100,13 +100,26 @@
              (when (ms/drained? scans-stream)
                (ms/close! scans-with-details)))))]
     (ms/consume-async add-details scans-stream)
+    (ms/on-drained scans-stream #(ms/close! scans-with-details))
```
derwolfe: Just to put a comment in for people looking at this in the future. I think this will work for now, but, there should be a way to communicate this without needing to register a handler function to perform cleanup. Using a function like `manifold.stream/connect-via` should be able to handle closing streams that are downstream of another stream; e.g. the stream sitting downstream of scans-stream, namely scans-with-details, would be notified automatically that `scans-stream` has been closed, and close itself, without our needing to register a handler. :-)

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: Done. Let me know if the new message isn't specific enough.

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: Done.

src/cloudpassage_lib/scans.clj
```diff
@@ -100,13 +100,28 @@
...
     (ms/consume-async add-details scans-stream)
+    ; TODO: Figure out a way to automatically close the stream this function
+    ;       returns without using the lower-level on-drained callback.
+    (ms/on-drained scans-stream #(ms/close! scans-with-details))
     scans-with-details))
 
-(defn fim-report!
-  "Get the current (recent) FIM report for a particular client."
-  [client-id client-secret]
+(defn report-for-module!
```
derwolfe: What would you think of making this private api?

src/cloudpassage_lib/scans.clj
```diff
@@ -100,13 +100,28 @@
              (when (ms/drained? scans-stream)
                (ms/close! scans-with-details)))))]
     (ms/consume-async add-details scans-stream)
+    ; TODO: Figure out a way to automatically close the stream this function
```
derwolfe: This is an enormous nit that may actually just be for the benefit of emacs users :-).

Typically, `;` comments occur following a piece of code, e.g. `(ms/consume-async add-details scans-stream) ;blah blah`, and `;;` comments can occur on their own lines.

Emacs, by nature of being an extremely helpful, wonderful operating system, knows about this rule and auto-indents these comments. So, if a `;` is dangling on its own line, it will indent according to the previous line. Which can be a bit annoying :-).

My long-winded comment has really been a way for me to hide my shameful request: could you please change all of your new `;` comments to use `;;` if they occur on their own lines?

TimSimpsonR: I'll do it, even though comments like that scare me due to their resemblance to spider faces.

lvh: `//;;\\`

src/cloudpassage_lib/scans.clj
TimSimpsonR: Sure.

src/cloudpassage_lib/scans.clj
TimSimpsonR: @lvh Precisely.

src/cloudpassage_lib/scans.clj
```diff
@@ -100,13 +100,28 @@
...
     scans-with-details))
 
-(defn fim-report!
-  "Get the current (recent) FIM report for a particular client."
-  [client-id client-secret]
+(defn ^:private report-for-module!
+  "Get recent report data for a certain client, and ilter based on module."
+  [client-id client-secret module-name]
+  ; The docs say we can use "module" as a query parameter but it does
+  ; not work for FIM or SVM, so we have to filter out those items instead.
```
derwolfe: (this one too :smile: )

src/cloudpassage_lib/scans.clj
```diff
@@ -100,13 +100,28 @@
...
+    ;; TODO: Figure out a way to automatically close the stream this function
+    ;;       returns without using the lower-level on-drained callback.
+    (ms/on-drained scans-stream #(ms/close! scans-with-details))
     scans-with-details))
 
-(defn fim-report!
-  "Get the current (recent) FIM report for a particular client."
-  [client-id client-secret]
+(defn ^:private report-for-module!
+  "Get recent report data for a certain client, and ilter based on module."
```
ehashman: ilter -> filter

TimSimpsonR: Done.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -115,29 +115,44 @@
...
+    (let [empty-stream (ms/stream 0)]
+      (ms/close! empty-stream)
+      (->> empty-stream
+           (scans/scans-with-details! "client-id" "client-secret")
+           (ms/stream->seq)
+           doall))))
 
-(deftest fim-report!-tests
+(defn ^:private test-report
+  [report-func expected-module]
```
ehashman: I am not a fan of the arg name `report-func`; it seems a bit inconsistent. I think `report-fn` would be more consistent, but I think that calling it `report-for-module-type!` or the like would be a bit clearer when we invoke it below.

TimSimpsonR: This was a misspelling. I had intended to write report-funk.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -115,29 +115,44 @@
...
+                  :scan {}})
+               scans))
+        (is (ms/closed? scans-stream))
+        (is (ms/closed? scans-with-details)))))
+  (testing "Blank input stream won't block."
+    (let [empty-stream (ms/stream 0)]
+      (ms/close! empty-stream)
+      (->> empty-stream
+           (scans/scans-with-details! "client-id" "client-secret")
+           (ms/stream->seq)
```
ehashman: As lvh pointed out, there is no difference in the `->>` macro between invoking `ms/stream->seq` with and without parens, but to be consistent with our other use cases, can you remove them here?

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: I think adding an exclamation point makes sense. I think report-for-modulet-type! implies that the first argument should be the module, when here we want to keep calling the two different functions. On that note I kept around rather than making a single function that accepted a module parameter as I'm expecting SVM may deviate as we learn more going forward, though I think the case could be made we'd do better having a single function.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -115,29 +115,44 @@
...
+        (is (= (for [scan-id (range (* fake-pages scans-per-page))]
+                 {:scan-id scan-id
+                  :module (index->module scan-id)
+                  :url details-query-url
+                  :scan {}})
+               scans))
+        (is (ms/closed? scans-stream))
+        (is (ms/closed? scans-with-details)))))
+  (testing "Blank input stream won't block."
+    (let [empty-stream (ms/stream 0)]
```
ehashman: Even though we're not putting anything into this stream, testing a stream with a 0-item buffer seems weird to me. I don't think it will affect the test if we change this to match the stream with buffer size we're using in scans, so maybe you could do that?

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: Just to calrify I switched this to report-fn! though I'm open to other names.

ehashman: I suggested `report-for-module-type!` instead of just `report-for-module!` to remind the dev that the module type is actually bound in the function scope, although maybe that wasn't very clear.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -115,29 +115,44 @@
...
+      (ms/close! empty-stream)
+      (->> empty-stream
+           (scans/scans-with-details! "client-id" "client-secret")
```
ehashman: Small nit: since we don't actually care about the client-id/secret getting evaluated or being used anywhere, it can be nice to reflect that in what we're passing. In cloudpassage-reporter, I commonly pass `'_` in this case as arguments (e.g. `(fim-report! '_ '_)`).

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: It feels like a premature pessimization to create a buffer that won't be filled. It is a little weird, but I don't think a buffer size of zero could be effecting this test in other ways, since the root issue involves the output buffer which is created and returned by scans-with-details.

derwolfe: I also agree with tim here, I think having a non-zero sized buffer would be more distracting than informative.

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: Nice. I was actually wanting to do something similar but didn't know what the idiom was in Clojure, thanks. On a related note, I was tempted to create named bindings for the values "lvh" and "hunter" since it isn't clear when I read the code those are test values for client-id and client-secret, but didn't know if it would be expanding the scope of the PR.

test/cloudpassage_lib/scans_test.clj
ehashman: Mmk. I just like tests to reflect the contexts we actually use things in, but it sounds like I'm being overruled :)

ehashman:
Modulo that last outstanding comment, I poked around with this in the REPL and LGTM! We'll need to cut a new release once this lands.

PR #22, opened 2016-03-16
Use API-level filtering

ehashman:
Sorry in advance to Travis, this will break the tests.


ehashman:
I think this will end up being superseded by Tim's changes.

Issue #20, opened 2016-03-16
Add CSM reporting capability


ehashman:
Although "CSM" might actually be an inaccurate: it looks like the docs call this "SCA"? See https://support.cloudpassage.com/entries/24082902-Scan-History for more details.

ehashman:
Closed by #23.

PR #23, opened 2016-03-16
Adds csm-report! function

TimSimpsonR:


src/cloudpassage_lib/scans.clj
```diff
@@ -125,3 +125,8 @@
   "Get the current (recent) SVM report for a particular client."
   [client-id client-secret]
   (report-for-module! client-id client-secret "svm"))
+
+(defn csm-report!
```
derwolfe: Based on the other report functions it seems like this would be named`sca-report!` to match the naming from cloudpassage's api? I think that the pattern this library was using for function names was to use the cloudpassage name for the function, and the rackspace name to display the information in the templates.

TimSimpsonR:
Presumably solves 
"Add CSM reporting capability #20"

src/cloudpassage_lib/scans.clj
TimSimpsonR: I thought so too, but believed you and @ehashman were asking that the names match our names. But I think using the Cloud Passage API words here makes sense.

derwolfe: I must have misunderstood, sorry about that. My understanding was that our library would be exposing the names as they appear from cloudpassage. :-).

ehashman: My understanding is the same as Chris, we should be treating CloudPassage as authoritative and conform to their iron will.

TimSimpsonR: Updating based on your comments.

derwolfe:
LGTM - @ehashman?

ehashman:
+1

PR #24, opened 2016-03-17
Off by one error

derwolfe:
pagination off by one page?

### DIFF
--- a/src/cloudpassage_lib/scans.clj
+++ b/src/cloudpassage_lib/scans.clj
@@ -71,12 +71,12 @@
                   (fn [response]
                     (let [{:keys [pagination scans]} response
                           next-url (:next pagination)]
+                      (ms/put-all! scans-stream scans)
                       (if (or (= "" next-url) (nil? next-url))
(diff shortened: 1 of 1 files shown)


TimSimpsonR:
If I’m reading it right, it looks like the if statement in scans.clj on line 74 could be closing scans-stream even though line 79 pushes to it. So moving line 79 to before line 74 would add one more page of results.

TimSimpsonR:
Really glad I could leave the insightful comment. I was following this in #eng and thought this was an issue, not a PR to fix it...  

+1

derwolfe:
@TimSimpsonR - sorry, for making the unclear PR. I just wanted to make the CI fail to display this issue more clearly (otherwise I couldn't get a reproduction of the error using an issue alone). I'm not currently working on fixing this.

TimSimpsonR:
@derwolfe No, nothing wrong on your end, we had just been talking about this problem in #eng and when I saw the message for the GitHub link I assumed it was an issue, not a PR as the conversation suggested that uncertainty as to where the bug was.

PR #28, opened 2016-03-17
Convert all report responses to kebab-case

derwolfe:
This will be a breaking change for api consumers. 

I've chosen to perform this step inside of `report-for-module` because it needs to occur for *all* data returned by the api; `report-for-module` is responsible for producing all of this data.


derwolfe:
Ugh - pushed the wrong branch.

PR #29, opened 2016-03-17
Convert all report responses to kebab-case

derwolfe:
This change will affect consumers depending on snake-case being present.


test/cloudpassage_lib/scans_test.clj
```diff
@@ -159,4 +178,3 @@
 
 (deftest sca-report!-tests
   (test-report scans/sca-report! "sca"))
-
```
ehashman: eheheh +1

ehashman:
This LGTM, let me just pull it down in a REPL and poke around a bit before I merge.

PR #30, opened 2016-03-17
Change API calls used to get scan data

TimSimpsonR:
Previously we've been using the scans API to fetch data by time to
generate our report. This PR changes it to fetch the most recent
scans for each server (aka host) which keeps us from having to
use large time periods in our query to make sure we get some data
for each server while also having to throw our duplicate info for
any server which may appear.

### DIFF
--- a/src/cloudpassage_lib/scans.clj
+++ b/src/cloudpassage_lib/scans.clj
@@ -23,6 +23,9 @@
 (def ^:private scans-path
   (partial str "/v1/scans/"))
 
+(def ^:private base-servers-url
+  "https://api.cloudpassage.com/v1/servers/")
+
 (defn ^:private maybe-flatten-list
   [maybe-list]
   (if (or (string? maybe-list) (nil? maybe-list))
@@ -37,6 +40,11 @@
      (-> (u/url url)
          (update :query merge opts)
          str))))
+
+(defn ^:private scan-server-url
+  "URL for fetching most recent scan results of a server."
+  [server-id module]
+  (str (u/url "https://api.cloudpassage.com/v1/servers/" server-id module)))
 
 (defn ^:private scans-detail-url
   "Compute the URL for a scan detail."
@@ -94,6 +102,11 @@
   [client-id client-secret opts]
   (stream-paginated-resources! client-id client-secret (scans-url opts) :scans))
 
+(defn list-servers!
+  "Returns a stream of servers for the given account."
+  [client-id client-secret]
+  (stream-paginated-resources! client-id client-secret base-servers-url :servers))
+
 (defn scans-with-details!
   "Returns a stream of historical scan results with their details.
 
@@ -114,17 +127,36 @@
      scans-with-details-stream)
     scans-with-details-stream))
 
+(defn scan-server
+  [client-id client-secret server-id module]
+  (let [url (scan-server-url server-id module)]
+    (get-page! client-id client-secret url)))
+
+(defn scan-each-server!
+  "Given a stream of servers, returns a stream of scan data for each server."
+  [client-id client-secret module servers-stream]
+  (let [server-details-stream (ms/stream 10)]
+    (ms/connect-via
+     servers-stream
+     (fn [{:keys [id]}]
+       (md/chain
+        (scan-server client-id client-secret id module)
+        (fn [response]
+          (if (cpc/page-response-ok? response)
+            (ms/put! server-details-stream response)
+            (error "Error getting scans for server " id)))))
+     server-details-stream)
+    server-details-stream))
+
 (defn ^:private report-for-module!
   "Get recent report data for a certain client, and filter based on module."
   [client-id client-secret module-name]
   ;; The docs say we can use "module" as a query parameter but it does
   ;; not work for FIM or SVM, so we have to filter out those items instead.
-  (let [opts {"since" (cpc/->cp-date (-> 3 hours ago))}]
-    (->> (scans! client-id client-secret opts)
-         (ms/filter (fn [{:keys [module]}] (= module module-name)))
-         (scans-with-details! client-id client-secret)
-         (ms/map #(cske/transform-keys cskc/->kebab-case-keyword %))
-         ms/stream->seq)))
+  (->> (list-servers! client-id client-secret)
+       (scan-each-server! client-id client-secret module-name)
+       (ms/map #(cske/transform-keys cskc/->kebab-case-keyword %))
+       ms/stream->seq))
 
 (defn fim-report!
   "Get the current (recent) FIM report for a particular client."
--- a/test/cloudpassage_lib/scans_test.clj
+++ b/test/cloudpassage_lib/scans_test.clj
@@ -44,6 +44,10 @@
   (is (= "https://abc.com/v1/scans/abcdef/findings/xyzzy"
          (#'scans/finding-detail-url "https://abc.com/" "abcdef" "xyzzy"))))
 
+(deftest scan-server-url-tests
+  (is (= "https://api.cloudpassage.com/v1/servers/server-id/svm"
+         (#'scans/scan-server-url "server-id" "svm"))))
+
 (defn ^:private index->module
   "Given an index of a (fake, test-only) scan, return a module for that scan."
   [i]
@@ -163,28 +167,112 @@
            ms/stream->seq
            doall))))
 
+(defn ^:private mock-get-page
+  "Generates a fake page function. Give it the expected base path and a
+   function to call with the current page number."
+  [cb]
+  (fn [client-id client-secret url]
+    (is (= client-id "lvh"))
+    (is (= client-secret "hunter2"))
+    (let [parsed-url (u/url url)
+          path (:path parsed-url)
+          query (:query parsed-url)
+          page-num (-> query
+                       (get "page" "0")
+                       Integer/parseInt)]
+      (cb {:page-num page-num :path path}))))
+
+(deftest list-servers!-tests
+  (testing "Returns all servers if paginated call is OK."
+    (println #'scans/base-servers-url)
+    (with-redefs [scans/get-page!
+                  (mock-get-page
+                   (fn [{:keys [page-num path]}]
+                     (is (= "/v1/servers" path))
+                     (case page-num
+                       0 {:servers [{:id "server-id-1"} {:id "server-id-2"}]
+                          :pagination {:next (str @#'scans/base-servers-url "?page=1")}}
+                       1 {:servers [{:id "server-id-3"}]})))]
+      (let [server-stream (scans/list-servers! "lvh" "hunter2")
+            id-list (map :id (ms/stream->seq server-stream))]
+        (is (= ["server-id-1" "server-id-2" "server-id-3"] id-list))
+        (is (ms/closed? server-stream))))
+    (testing "If page 2 is bad, a partial list is returned."
+      (with-redefs [scans/get-page!
+                    (mock-get-page
+                     (fn [{:keys [page-num path]}]
+                       (is (= "/v1/servers" path))
+                       (case page-num
+                         0 {:servers [{:id "server-id-1"} {:id "server-id-2"}]
+                            :pagination {:next (str @#'scans/base-servers-url "?page=1")}}
+                         1 :cloudpassage-lib.core/fetch-error)))]
+        (let [server-stream (scans/list-servers! "lvh" "hunter2")
+              id-list (map :id (ms/stream->seq server-stream))]
+          (is (= ["server-id-1" "server-id-2"] id-list))
+          (is (ms/closed? server-stream)))))))
+
+(deftest scan-server-tests
+  (with-redefs
+   [scans/get-page! (mock-get-page
+                     (fn [{:keys [path]}]
+                       (is (= "/v1/servers/server-id-here/svm" path))
+                       "GOOD"))]
+    (is (= "GOOD" (scans/scan-server "lvh" "hunter2" "server-id-here" "svm")))))
+
+(deftest scan-each-server!-tests
+  (testing "Scan can handle an empty stream."
+    (let [input (ms/stream)]
+      (ms/close! input)
+      (let [scan-stream (scans/scan-each-server! "lvh" "hunter2" "svm"
+                                                 input)
+            scan-result (clojure.string/join "" (ms/stream->seq scan-stream))]
+        (is (= "" scan-result)))))
+  (testing "When the server scan page is broken, we get back nadda."
+    (with-redefs [scans/get-page! (mock-get-page
+                                   (fn [{:keys [path]}]
+                                     (is (= "/v1/servers/server-id-here/svm" path))
+                                     :cloudpassage-lib.core/fetch-error))]
+      (let [input (ms/stream)]
+        (ms/put! input {:id "server-id-here"})
+        (ms/close! input)
+        (let [scan-stream (scans/scan-each-server! "lvh" "hunter2" "svm"
+                                                   input)
+              scan-result (clojure.string/join "" (ms/stream->seq scan-stream))]
+          (is (= "" scan-result))))))
+  (testing "Server scan can get back several pages of data."
+    (with-redefs [scans/get-page! (mock-get-page
+                                   (fn [{:keys [path]}]
+        ;; The real data returning from a valid scan-server call would be a
+        ;; map containing a bunch of scan data. But for testing purposes all
+        ;; that matters is that scan-each-server! puts whatever comes back
+        ;; into the its output stream.
+                                     (case path
+                                       "/v1/servers/server-1/svm" "ONE"
+                                       "/v1/servers/server-2/svm" "TWO"
+                                       "/v1/servers/server-3/svm" "THREE!")))]
+      (let [input (ms/stream 2)]
+        (ms/put-all! input [{:id "server-1"} {:id "server-2"} {:id "server-3"}])
+        (ms/close! input)
+        (let [scan-stream
+              (scans/scan-each-server! "lvh" "hunter2" "svm" input)
+              scan-result
+              (clojure.string/join "..." (ms/stream->seq scan-stream))]
+          (is (= "ONE...TWO...THREE!" scan-result)))))))
+
 (defn ^:private test-report
   [report-fn! expected-module]
-  (with-redefs [scans/get-page! fake-get-page!]
-    (let [report (report-fn! "lvh" "hunter2")]
-      (is (= (for [scan-id (range (* fake-pages scans-per-page))
-                   :let [module (index->module scan-id)]
-                   :when (= module expected-module)]
-               {:scan-id scan-id
-                :module module
-                :url details-query-url
-                :scan {}})
-             report))))
-  (with-redefs [scans/get-page! fake-get-page-with-snakes]
-    (let [report (report-fn! "lvh" "hunter2")]
-      (is (= (for [scan-id (range (* fake-pages scans-per-page))
-                   :let [module (index->module scan-id)]
-                   :when (= module expected-module)]
-               {:scan-id scan-id
-                :module module
-                :url details-query-url
-                :scan {}})
-             report)))))
+  (with-redefs
+   [scans/get-page!
+    (mock-get-page
+     (fn [{:keys [path]}]
+       (cond
+         (= path "/v1/servers") {:servers [{:id "server-id-1"} {:id "server-id-2"}]}
+         (= path (str "/v1/servers/server-id-1/" expected-module)) {:id "1" :scan {}}
+         (= path (str "/v1/servers/server-id-2/" expected-module)) {:id "2" :scan {}})))]
+    (let [report (report-fn! "lvh" "hunter2")
+          result (clojure.string/join "..." report)]
+      (is (= (str {:id "1", :scan {}} "..." {:id "2", :scan {}})
+             result)))))
 
 (deftest fim-report!-tests
   (test-report scans/fim-report! "fim"))


TimSimpsonR:
@derwolfe @ehashman This should be ready now. Please review.

src/cloudpassage_lib/scans.clj
```diff
@@ -9,11 +9,9 @@
    [manifold.stream :as ms]
    [environ.core :refer [env]]
    [cloudpassage-lib.core :as cpc]
-   [taoensso.timbre :as timbre :refer [info spy]]
+   [taoensso.timbre :as timbre :refer [error info spy]]
    [clj-time.core :as t :refer [hours ago]]
    [clj-time.format :as tf]
-   [camel-snake-kebab.core :as cskc]
-   [camel-snake-kebab.extras :as cske]
```
ehashman: Why is this stuff being removed? Perhaps you need to rebased on master?

src/cloudpassage_lib/scans.clj
```diff
@@ -37,6 +38,11 @@
          (update :query merge opts)
          str))))
 
+(defn ^:private scan-server-url
+  "URL for fetching most recent scan results of a server."
+  [server-id module]
+  (str (u/url "https://api.cloudpassage.com/v1/servers/" server-id module)))
```
ehashman: Rather than using str, you might want to consider writing this similar to `scans-url`? I did the same thing for building URLs in the tests until I saw that code.

src/cloudpassage_lib/scans.clj
```diff
@@ -61,28 +67,63 @@
   (let [token (cpc/fetch-token! client-id client-secret (:fernet-key env))]
     (cpc/get-single-events-page! token url)))
 
+(defn page-response-ok?
+  [response]
+  (not= :cloudpassage-lib.core/fetch-error response))
```
ehashman: Yay error handling! Much +1! But perhaps this should be split out into a separate PR?

src/cloudpassage_lib/scans.clj
TimSimpsonR: Because I removed the old report-for-module! method it looks like cljfmt took this out. 

I'm a bit confused- are we changing the output of the CP API so that all of the objects are in kebab case? I read the chatter yesterday but thought you and @derwolfe were discussing changing code, not output from the API.

src/cloudpassage_lib/scans.clj
```diff
@@ -61,28 +67,63 @@
...
+                          (close!)
+                          (Exception. "Error fetching scans."))))))]
     (ms/put! urls-stream (scans-url opts))
     (ms/consume-async shovel urls-stream)
     scans-stream))
 
+(defn list-servers!
+  "Returns a stream of servers for the given account."
+  [client-id client-secret]
+  (let [urls-stream (ms/stream 10) ;; absolutely no science here
```
ehashman: This code appears to have been copy-pasted from `scans!` above. I wonder if there is a way to refactor it for better reuse.

src/cloudpassage_lib/scans.clj
TimSimpsonR: Doesn't scans-url use str?

src/cloudpassage_lib/scans.clj
ehashman: Yes, we changed the stuff returned from the API from snake case to kebab case to be more stylistically consistent. That hasn't been released yet, but since it's a breaking change we'll probably do a major release. There's an associated issue on the reporter side to integrate the case changes, I think @derwolfe said he might work on that today if he had time.

src/cloudpassage_lib/scans.clj
ehashman: I haven't looked at the body, but it might be something that could be refactored to use various base URLs and arguments. (Hard to go check it on mobile, sorry)

src/cloudpassage_lib/scans.clj
```diff
@@ -107,16 +148,51 @@
     (ms/on-drained scans-stream #(ms/close! scans-with-details))
     scans-with-details))
 
+(defn scan-server
+  [client-id client-secret server-id module]
+  (let [url (scan-server-url server-id module)]
+    (get-page! client-id client-secret (scan-server-url server-id module))))
```
ehashman: I think this line meant to consume the URL in the let, but that got missed.

src/cloudpassage_lib/scans.clj
TimSimpsonR: It looks like it was only being used in the report-for-module! function. Should we be doing this for the individual API calls instead?

src/cloudpassage_lib/scans.clj
TimSimpsonR: scans-url looks like this:
```clojure
(-> (u/url url)
         (update :query merge opts)
         str)
```
Since there was no need to use update, using -> seemed like overkill.

src/cloudpassage_lib/scans.clj
TimSimpsonR: Ah yes. Thanks for the catch.

src/cloudpassage_lib/scans.clj
```diff
@@ -61,28 +67,63 @@
...
+                    (if (page-response-ok? response)                      
+                      (let [{:keys [pagination servers]} response
+                            next-url (:next pagination)]
+                        (ms/put-all! servers-stream servers)
+                        (if (or (= "" next-url) (nil? next-url))
+                          (do (info "end of servers pagination")
+                              (close!))
+                          (ms/put! urls-stream next-url)))
+                      (do (error "Error fetching server list for url:" url)
+                          (close!))))))]
```
derwolfe: What are these calls to `close!` closing?

src/cloudpassage_lib/scans.clj
TimSimpsonR: I added this because it made it easier to diagnose some issues I was seeing. I think it makes sense to go ahead and commit it.

src/cloudpassage_lib/scans.clj
derwolfe: Ah - never mind my question, this is new function with the same name.

src/cloudpassage_lib/scans.clj
TimSimpsonR: You're right, and actually the similarities have been bothering me. I'll go ahead and add a common function now. :)

src/cloudpassage_lib/scans.clj
ehashman: I think that's a presentation layer thing, so it makes a lot of sense to keep it at the reporting level.

TimSimpsonR: Ok- so I'm reading that as we're putting the snake-kebab in the reporting project and taking it out of here?

ehashman: Negative, I mean reporting layer on the API, the `foo-report!` stuff. Probably should have specified that.

src/cloudpassage_lib/scans.clj
```diff
@@ -107,16 +148,51 @@
...
+                 (when (ms/drained? input)
+                   (ms/close! output)))
+               (error "Error getting scans for server " id)))))]
+    (ms/consume-async fetch-server-details input)
+    ;; TODO: Figure out a way to automatically close the stream this function
+    ;;       returns without using the lower-level on-drained callback.
+    (ms/on-drained input #(ms/close! output))
+    output))
+
+; (defn ^:private report-for-module!
```
ehashman: Need to remove this commented code, which appears to be copied below anyways :)

src/cloudpassage_lib/scans.clj
```diff
@@ -107,16 +148,51 @@
...
+;          (ms/filter (fn [{:keys [module]}] (= module module-name)))   
+;          (scans-with-details! client-id client-secret)
+;          ms/stream->seq)))
+
 (defn ^:private report-for-module!
   "Get recent report data for a certain client, and filter based on module."
   [client-id client-secret module-name]
   ;; The docs say we can use "module" as a query parameter but it does
   ;; not work for FIM or SVM, so we have to filter out those items instead.
   (let [opts {"since" (cpc/->cp-date (-> 3 hours ago))}]
```
ehashman: Opts are no longer getting used anywhere

src/cloudpassage_lib/scans.clj
```diff
@@ -107,16 +148,51 @@
...
-    (->> (scans! client-id client-secret opts)
-         (ms/filter (fn [{:keys [module]}] (= module module-name)))
-         (scans-with-details! client-id client-secret)
-         (ms/map #(cske/transform-keys cskc/->kebab-case-keyword %))
```
ehashman: Just to reiterate, I am advocating for the status quo on the kebab case stuff. This should be left in here as it is IMO.

src/cloudpassage_lib/scans.clj
TimSimpsonR: Thanks- removed.

src/cloudpassage_lib/scans.clj
```diff
@@ -107,16 +148,51 @@
...
+    (->> (list-servers! client-id client-secret)
+         (scan-each-server! client-id client-secret module-name)
```
ehashman: Won't `scan-each-filter!` also need to be followed with `scans-with-details!`? It should be returning the same data as `scans!`, no?

src/cloudpassage_lib/scans.clj
TimSimpsonR: Ok- looking at it it's just different enough from scans that making a common function is tough.

Also, every time I try to refactor this I go into total yak-shaving mode because I think in general the way we're constructing these output streams in functions that accept input streams is possibly wrong or subject to race conditions. 

Maybe I'm looking at things incorrectly but here's where I see a possible race.

* The input stream has on-drained registered to a function that closes the output stream. This is so something will close the output stream if the input stream is empty and it won't block forever. However that means it could close the output stream before the deferred has a chance to write to it.

* Because the consume-async is calling a callback function that generates deferreds via md/chain, it could have multiple of these in flight waiting for a response as they call the CP API. However if one of them detects in the response that the input stream is empty, it will close the shared output stream, meaning the other deferreds won't be able to write to it.

Does that sound right or am I looking at this wrong?

src/cloudpassage_lib/scans.clj
TimSimpsonR: I dunno. ¯\_(ツ)_/¯

src/cloudpassage_lib/scans.clj
TimSimpsonR: Why would only this function need to have kebab case enforced? Seems like its a filter we'd want to apply in the report project as the intent is to make the templating more intuitive there.

src/cloudpassage_lib/scans.clj
TimSimpsonR: Ok- I actually was looking through the Manifold Streams source, and tried a few things before landing on using connect-via... which is what @derwolfe and @lvh had told me to try days ago, but for some reason had not stopped the function from blocking on an empty string. For whatever reason though it now works.

After changing which removed the need to check for and close the output stream creating a common function for scan-each-server! and scans-with-details! was pretty easy.

src/cloudpassage_lib/scans.clj
ehashman: If Clojure code is going to consume the data, we want the data to be spit out in kebab case, since that is idiomatic. If other non-Clojure code is going to consume from this API then I'd say put it at the reporting level.

src/cloudpassage_lib/scans.clj
TimSimpsonR: Ok, so I went into the reporter, changed it's project.clj to use the newest snap shot, did a local lein install, etc so I could test.

Turns out this changes the report from being a series of maps with scan data to a series of maps with server data that has the scan data available if you simply access the :scan key. 

In other words, the report works if I change:
```clojure
(let [num-hosts (count report)
        num-crit-hosts (-> (filter #(> (:critical_findings_count %) 0) report)
                           count)
        percent-crit (if (> num-hosts 0)
                       (* (/ num-crit-hosts num-hosts) 100)
                       0)
        num-crit (reduce + (map :critical_findings_count report))
        num-noncrit (reduce + (map :non_critical_findings_count report))
```
to:
```clojure
(let [num-hosts (count report)
        num-crit-hosts (-> (filter #(> (:critical_findings_count (:scan %)) 0) report)
                           count)
        percent-crit (if (> num-hosts 0)
                       (* (/ num-crit-hosts num-hosts) 100)
                       0)
        num-crit (reduce + (map #(:critical_findings_count (:scan %)) report))
        num-noncrit (reduce + (map #(:non_critical_findings_count (:scan %)) report))
```

TimSimpsonR: @derwolfe @ehashman I've emailed you both the generated report.

TimSimpsonR:
@derwolfe @ehashman I believe everything brought up here today has been addressed. Can you give it another look over?

src/cloudpassage_lib/scans.clj
ehashman: Could we not just pass in `shovel` as an argument to some more general function shaped like this? I feel like there were some local bindings that were a problem which is why we didn't do that before, but I think we can take care of that with dynamic scoping?

TimSimpsonR: I wrote the map-stream function which accepts a call back function. When I tried to rewrite this to use map, which would have moved shovel to an anonymous function and made it much simpler looking, it blocked endlessly while running the tests. I'm not sure what the culprit was other than the fact that this and scans! both write to their own input stream as well as the output stream. Still think it should be rewritten, I just haven't been able to do it yet without it breaking.

lvh: When using mine and Chris' suggestion to first abstract away a function that simply returns a stream of request bodies (walking the tree); you can quite easily do the actually specific parts + custom error handling with a simple `(map :scans)` or similar. There should be only one place that still has a `shovel`, so getting it out becomes a moot point. You could solve this with a dynamically scoped variable, but it seems like that variable would just increase the scope of some mutable state; so I don't think this is an example of a particularly good use case for `^:dynamic`.

lvh: (There's also the matter that that will make the change to use `connect-via` or `transform`, which Tim hinted at,  a lot hairier.)

TimSimpsonR: @lvh @ehashman I just pushed a commit that changes both of these to use map-stream.

However, I'm seeing some zaniness here. I'll comment on it in line to make it easier to follow.

src/cloudpassage_lib/scans.clj
```diff
@@ -61,27 +70,63 @@
...
+                                 (do (info "URL=" url) (get-page! client-id client-secret url))
+                                 (fn [response]
+                                   (if (page-response-ok? response)
+                                     (let [{:keys [pagination servers]} response
+                                           next-url (:next pagination)]
+                                       (if (blank? next-url)
+                                         (do (info "end of servers pagination")
+                                             (ms/close! urls-stream))
+                                         (ms/put! urls-stream next-url))
+                                       (ms/put-all! servers-stream servers))
```
TimSimpsonR: @lvh @ehashman So this is what I changed.

Btw, I figured out what was tripping me up. If ms/put-all! isn't returned from the function, it doesn't get fully executed. I'd moved the call to before the if statement which of course was no bueno.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -94,7 +98,7 @@
         page-num (-> query
                      (get "next" "0")
                      Integer/parseInt)
-        next-page (if (< page-num fake-pages)
+        next-page (if (< page-num (- fake-pages 1))
```
TimSimpsonR: @ehashman @lvh Want to point out that I had to change this to make the tests pass.

For some reason when I changed scans! to use map-stream, I was getting test failures. In the process I cranked the test pages and scans per page down from 10 and 10 to 3 and 5 to see what was happening and make the diffs smaller.

Turns out that though the fake pages should have always been returning fake-pages + 1 in total, but for some reason the test had been passing anyway. Near as I can tell it was a bugs in the scans code getting less than the full number of pages, as when I looked at it I couldn't figure out how fake-pages could ever have been the exactly the number of full pages returned by scans without the change I implemented above.

src/cloudpassage_lib/scans.clj
```diff
@@ -61,27 +70,63 @@
...
+                                     (do (error "Error fetching server list for url:" url)
+                                         (ms/close! urls-stream)
+                                         (Exception. "Error fetching servers."))))))))))
```
TimSimpsonR: I feel like cljfmt is making this code look a lot uglier than it did when I wrote it. This is what the formatting was like originally:

```clojure
(defn list-servers!
  "Returns a stream of servers for the given account."
  [client-id client-secret]
  (let [urls-stream (ms/stream 10)]
    (ms/put! urls-stream base-servers-url)
    (map-stream urls-stream (fn [servers-stream]
      (fn [url]
      (md/chain
        (do (info "URL=" url) (get-page! client-id client-secret url))
        (fn [response]
          (if (page-response-ok? response)
            (let [{:keys [pagination servers]} response
                  next-url (:next pagination)]
              (if (blank? next-url)
                (do (info "end of servers pagination")
                    (ms/close! urls-stream))
                (ms/put! urls-stream next-url))
              (ms/put-all! servers-stream servers))
            (do (error "Error fetching server list for url:" url)
              (ms/close! urls-stream)
              (Exception. "Error fetching servers."))))))))))
```

TimSimpsonR: If I move the arguments on map-stream to one per line,cljfmt gives me this less horizontally stretched version. Is this generally what I should do to avoid making the code appear in Cinemascope? 

```clojure
(defn list-servers!
  "Returns a stream of servers for the given account."
  [client-id client-secret]
  (let [urls-stream (ms/stream 10)]
    (ms/put! urls-stream base-servers-url)
    (map-stream
     urls-stream
     (fn [servers-stream]
       (fn [url]
         (md/chain
          (do (info "URL=" url) (get-page! client-id client-secret url))
          (fn [response]
            (if (page-response-ok? response)
              (let [{:keys [pagination servers]} response
                    next-url (:next pagination)]
                (if (blank? next-url)
                  (do (info "end of servers pagination")
                      (ms/close! urls-stream))
                  (ms/put! urls-stream next-url))
                (ms/put-all! servers-stream servers))
              (do (error "Error fetching server list for url:" url)
                  (ms/close! urls-stream)
                  (Exception. "Error fetching servers."))))))))))
```

derwolfe: Yeah, this would be preferable to the previous version. That being said, I'd probably like to to see the function you're using here defined outside of `(map-stream... )`. But, I can see the argument for *not* doing this (why give a name to a thing that is only used the single time :-).

TimSimpsonR: I went ahead and made a common function for scans! and list-servers! First I tried tot extract the inner function, but when I looked at it making a common function for scans! and list-servers! was less noisy.

src/cloudpassage_lib/scans.clj
```diff
@@ -61,27 +70,54 @@
...
+  "Maps an input stream to an output stream with some function.
+
+  The function is expected to accept the output stream and return another
+  function that is called repeatedly for each bit of input."
+  [input f]
+  (let [output (ms/stream)]
+    (ms/connect-via input (f output) output)
+    output))
+
+(defn paginated-list!
```
derwolfe: What would you think of marking this as private?

src/cloudpassage_lib/scans.clj
```diff
@@ -61,27 +70,54 @@
   (let [token (cpc/fetch-token! client-id client-secret (:fernet-key env))]
     (cpc/get-single-events-page! token url)))
 
+(defn page-response-ok?
+  [response]
+  (not= :cloudpassage-lib.core/fetch-error response))
+
+(defn map-stream
```
derwolfe: Would you be ok with marking this as private?

test/cloudpassage_lib/scans_test.clj
```diff
@@ -112,16 +116,32 @@
...
+    (with-redefs [scans/get-page! fake-get-page!]
+      (let [scans-stream (scans/scans! "lvh" "hunter2" {"modules" "fim"})
+            scans (ms/stream->seq scans-stream)]
+        (is (= (for [scan-id (range (* fake-pages scans-per-page))]
+                 {:scan-id scan-id
+                  :module (index->module scan-id)
+                  :url details-query-url})
+               scans))
+        (is (ms/closed? scans-stream)))))
+  (testing "Error occurs fetching page."
```
derwolfe: Would you mind changing this testing declaration to describe what is happening, e.g "On error an empty result is returned" or something similar?

src/cloudpassage_lib/scans.clj
ehashman: I had to look at the docstring to get an idea of what `paginated-list!` did. Perhaps we could use a more descriptive name, like `get-paginated-resources!` (just a suggestion based on your docstring below)?

test/cloudpassage_lib/scans_test.clj
```diff
@@ -147,28 +167,117 @@
...
+                    (is (= "/v1/servers" path))
+                    (case page-num
+                      0 {:servers [{:id "server-id-1"} {:id "server-id-2"}]
+                         :pagination {:next (str @#'scans/base-servers-url "?page=1")}}
+                      1 (if page-two-breaks
+                          :cloudpassage-lib.core/fetch-error
+                          {:servers [{:id "server-id-3"}]})))))
+
+(deftest list-servers!-tests
+  (testing "Make sure the server call returns some servers."
```
derwolfe: I think you can safely omit "make sure the..." and say only "List-servers returns a list of servers" or the like. :-).

src/cloudpassage_lib/scans.clj
```diff
@@ -61,27 +70,54 @@
...
+    output))
+
+(defn paginated-list!
+  "Returns a stream of resources coming from a paginated list."
+  [client-id client-secret initial-url resource-key]
+  (let [urls-stream (ms/stream 10)]
+    (ms/put! urls-stream initial-url)
+    (map-stream
+     urls-stream
+     (fn [scans-stream]
```
ehashman: Can we possibly factor out this lambda into the let like in `scans!`, for easier reading?

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: Sure.

src/cloudpassage_lib/scans.clj
TimSimpsonR: I tried it but the function would need to get passed client-id, client-secret, url, urls-stream, and scans-stream as arguments. Here's what it looks like:

```clojure
(defn ^:private get-paginated-page!
  [client-id client-secret resource-key urls-stream scans-stream url]
  (md/chain
    (get-page! client-id client-secret url)
    (fn [response]
      (if (page-response-ok? response)
        (let [resource (resource-key response)
              pagination (:pagination response)
              next-url (:next pagination)]
          (if (blank? next-url)
            (do (info "no more urls to fetch")
                (ms/close! urls-stream))
            (ms/put! urls-stream next-url))
          (ms/put-all! scans-stream resource))
        (do (error "Error getting scans for url: " url)
            (ms/close! urls-stream)
            (Exception. "Error fetching scans."))))))

(defn ^:private paginated-list!
  "Returns a stream of resources coming from a paginated list."
  [client-id client-secret initial-url resource-key]
  (let [urls-stream (ms/stream 10)]
    (ms/put! urls-stream initial-url)
    (map-stream urls-stream
     (fn [scans-stream]
      #(get-paginated-page! client-id client-secret resource-key urls-stream scans-stream %)))))

```

Also this version closes the url-stream outside of the function it was created in, which makes it a smidge harder to track how that resource is being used IMO.

src/cloudpassage_lib/scans.clj
TimSimpsonR: That is a better name- though thinking about it "stream" might be better than "get" as it implies the return type. Changed.

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: Done.

TimSimpsonR:
@ehashman @derwolfe I believe I've updated everything. Let me know if I've missed something...

src/cloudpassage_lib/scans.clj
derwolfe: @TimSimpsonR - The following is a solution that might help readability, while still keeping all of the logic local to `stream-paginated-resources`:

```clojure
(defn ^:private stream-paginated-resources!
  "Returns a stream of resources coming from a paginated list."
  [client-id client-secret initial-url resource-key]
  (let [urls-stream (ms/stream 10)
        fetch-page (fn [url]
                     (md/chain
                      (get-page! client-id client-secret url)
                      (fn [response]
                        (if (page-response-ok? response)
                          (let [resource (resource-key response)
                                pagination (:pagination response)
                                next-url (:next pagination)]
                            (if (blank? next-url)
                              (do
                                (info "no more urls to fetch")
                                (ms/close! urls-stream))
                              (ms/put! urls-stream next-url))
                            (ms/put-all! scans-stream resource))
                          (do
                            (error "Error getting scans for url: " url)
                            (ms/close! urls-stream)
                            (Exception. "Error fetching scans."))))))]
    (ms/put! urls-stream initial-url)
    (map-stream urls-stream fetch-page)))
```
(this implemenation should work, but I've not executed any tests verifying its correctness :-) )

@ehashman - does this look more like what you were suggesting?

ehashman: Yes, although I think your fetch-page is missing one outer lambda wrapper that takes scans-stream as an argument.

TimSimpsonR: Got it... I'll do that in this PR: https://github.com/RackSec/cloudpassage-lib/pull/33

derwolfe: Drat - sorry, I must have had copy paste brain going. Thanks for pointing that out @ehashman

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,28 +167,117 @@
...
+        ;; The real data returning from a valid scan-server call would be a
+        ;; map containing a bunch of scan data. But for testing purposes all
+        ;; that matters is that scan-each-server! puts whatever comes back
+        ;; into the its output stream.
+                                     (case path
+                                       "/v1/servers/server-1/svm" "ONE"
+                                       "/v1/servers/server-2/svm" "TWO"
+                                       "/v1/servers/server-3/svm" "THREE!")))]
+      (let [input (ms/stream)]
+        (ms/put! input {:id "server-1"})
```
derwolfe: This could be replaced with `(ms/put-all! input [{:id "server-1"} {:id "server-2"} {:id "server-3"}])`

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,28 +167,117 @@
...
+        (ms/put! input {:id "server-2"})
+        (ms/put! input {:id "server-3"})
+        (ms/close! input)
+        (let [scan-stream
+              (scans/scan-each-server! "lvh" "hunter2" "svm"
+                                       input)
```
derwolfe: Would you mind putting input back on 257? :-)

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,28 +167,117 @@
...
+    (let [input (ms/stream)]
+      (ms/close! input)
+      (let [scan-stream (scans/scan-each-server! "lvh" "hunter2" "svm"
+                                                 input)
+            scan-result (clojure.string/join "" (ms/stream->seq scan-stream))]
+        (is (= "" scan-result)))))
+  (testing "When the server scan page is broken, we get back nadda."
+    (with-redefs [scans/get-page! (make-fake-page
+                                   (fn [{:keys [path]}]
+                                     (case path
```
derwolfe: It doesn't look like path or the case statement are used here.

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: Oh nice. Done.

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: Done.

src/cloudpassage_lib/scans.clj
ehashman: There's actually a bunch of this URL stuff that needs cleanup. I'm going to file an issue so we can deal with that later. It would be out of scope for this PR.

src/cloudpassage_lib/scans.clj
```diff
@@ -94,6 +102,11 @@
   [client-id client-secret opts]
   (stream-paginated-resources! client-id client-secret (scans-url opts) :scans))
 
+(defn list-servers!
+  "Returns a stream of servers for the given account."
+  [client-id client-secret]
+  (stream-paginated-resources! client-id client-secret base-servers-url :servers))
```
ehashman: This is just a small nit but it seems it would make more sense to put this with the rest of the server logic instead of in between scans! and scans-with-details!

src/cloudpassage_lib/scans.clj
```diff
@@ -114,17 +127,36 @@
      scans-with-details-stream)
     scans-with-details-stream))
 
+(defn scan-server
```
ehashman: Given that this is only used inside scan-each-server!, it might make sense to move this to a local binding inside that function.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,28 +167,113 @@
...
+  [cb]
+  (fn [client-id client-secret url]
+    (is (= client-id "lvh"))
+    (is (= client-secret "hunter2"))
+    (let [parsed-url (u/url url)
+          path (:path parsed-url)
+          query (:query parsed-url)
+          page-num (-> query
+                       (get "page" "0")
+                       Integer/parseInt)]
```
ehashman: A lot of this logic was copy-pasted from fake-get-page!; I think this should live inside there.

TimSimpsonR: My impression was fake-get-page! was handling two possible URLs, one for the normal scans page and the other for the details. I was concerned that adding extra conditional logic to make it work for the servers pages would make that function confusing.

Looks aside, I actually didn't create this function by copying and pasting fake-get-page! initially but after I needed to do pagination I ended up copying the top chunks of that function. 

How about we make fake-get-page! call use make-fake-page! so this bit of logic isn't duplicated?

ehashman: `fake-get-page!` initially handled one kind of URL, and was expanded to work for the details URLs as well. You can definitely expand it to work for the servers URLs. It currently uses an `if` because there's only two types of pages, but that can be easily switched to a `cond`. Check out what we did with factoring out the logic for `fake-scans-page`: https://github.com/RackSec/cloudpassage-lib/blob/53d6b751c4e661c65651d5fa69e708c737e4cdd9/test/cloudpassage_lib/scans_test.clj#L60

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,28 +167,113 @@
...
+      (cb {:page-num page-num :path path}))))
+
+(defn fake-servers-page
+  "Pretends to be the servers list page results."
+  [page-two-breaks]
+  (make-fake-page (fn [{:keys [page-num path]}]
+                    (is (= "/v1/servers" path))
+                    (case page-num
+                      0 {:servers [{:id "server-id-1"} {:id "server-id-2"}]
```
derwolfe: Is it possible for the cloudpassage api to return a page 0?

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,28 +167,113 @@
...
+    (with-redefs [scans/get-page! (fake-servers-page false)]
+      (let [server-stream (scans/list-servers! "lvh" "hunter2")
+            id-list (map :id (ms/stream->seq server-stream))]
+        (is (= ["server-id-1" "server-id-2" "server-id-3"] id-list))
+        (is (ms/closed? server-stream)))))
+  (testing "If page 2 is bad, a partial list is returned."
+    (with-redefs [scans/get-page! (fake-servers-page true)]
+      (let [server-stream (scans/list-servers! "lvh" "hunter2")
+            id-list (map :id (ms/stream->seq server-stream))]
+        (is (= ["server-id-1" "server-id-2"] id-list))
```
ehashman: Is this the behaviour we want? If we can't fetch all the pages, we won't be able to return a list of all the servers, so we probably want to throw an exception and die. I'm pretty sure that's what the code says we do, but it looks like we're not actually exercising that logic here.

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: When I wrote this on Friday, the unit tests were modeling it this way. I can change it...

src/cloudpassage_lib/scans.clj
```diff
@@ -114,17 +127,36 @@
...
+
+(defn scan-each-server!
+  "Given a stream of servers, returns a stream of scan data for each server."
+  [client-id client-secret module servers-stream]
+  (let [server-details-stream (ms/stream 10)]
+    (ms/connect-via
+     servers-stream
+     (fn [{:keys [id]}]
+       (md/chain
+        (scan-server client-id client-secret id module)
```
ehashman: We might want to leave a docs note here that this returns the full details-level scan, and hence we don't need to call scans-with-details.

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: I saw that. I assumed though that the function wasn't supposed to be a full test fake (i.e. completely mocked out the Cloud Passage API) and had just been expanded as it worked with the scans function. It appeared easier to me to put the expected mocked value next to each unit tests, as normally unit tests don't utilize full test fakes.

Currently five tests are calling this so I another question is it worth the effort to rewrite it.

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: IIRC previously this would just add a nil or something to the stream when it found the error so logging it seemed like an improvement. I'm not sure how to make an error happening in the scans/list-servers! call raise an exception at the call site, but if I find out I'll change it. When that happens I should also change the error handling on the code that was recently submitted in my other PR.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,28 +167,113 @@
            ms/stream->seq
            doall))))
 
+(defn ^:private make-fake-page
```
derwolfe: @TimSimpsonR - It looks like this function is designed specifically so that you can reuse some of your assertion logic while also being able to pass in a function to perform further behavior. It seems like we could add a few more cases to `fake-get-page` to accomplish this, as long as they were positive behaviors - e.g. the query string would be able to drive the response. 

The concern I have with this design is that it makes the mocking a little bit harder to track (at least for me). Instead of being able to look directly at the function definition created for the specific test case OR at a single function containing all of the possible behaviors, one has to dig a bit deeper to understand what behavior is actually being mocked out.

Could you look into how we might be able to structure this a bit differently such that we aren't passing the callback that contains most of the behavior into the function?

ehashman:
Superseded by #35, which has been merged.

PR #32, opened 2016-03-18
update release notes for 2.0

derwolfe:


derwolfe:
Note - when this is released, the command to use will be `lein release :major`

PR #33, opened 2016-03-21
Fixes off by one bug, adds error handling

TimSimpsonR:
This fixes the off by one bug and adds error handling to the
scan! function.

### DIFF
--- a/src/cloudpassage_lib/scans.clj
+++ b/src/cloudpassage_lib/scans.clj
@@ -9,12 +9,13 @@
    [manifold.stream :as ms]
    [environ.core :refer [env]]
    [cloudpassage-lib.core :as cpc]
-   [taoensso.timbre :as timbre :refer [info spy]]
+   [taoensso.timbre :as timbre :refer [error info spy]]
    [clj-time.core :as t :refer [hours ago]]
    [clj-time.format :as tf]
    [camel-snake-kebab.core :as cskc]
    [camel-snake-kebab.extras :as cske]
-   [clojure.java.io :as io]))
+   [clojure.java.io :as io]
+   [clojure.string :refer [blank?]]))
 
 (def ^:private base-scans-url
   "https://api.cloudpassage.com/v1/scans/")
@@ -61,27 +62,37 @@
   (let [token (cpc/fetch-token! client-id client-secret (:fernet-key env))]
     (cpc/get-single-events-page! token url)))
 
+(defn ^:private stream-paginated-resources!
+  "Returns a stream of resources coming from a paginated list."
+  [client-id client-secret initial-url resource-key]
+  (let [urls-stream (ms/stream 10)
+        resources-stream (ms/stream 20)]
+    (ms/put! urls-stream initial-url)
+    (ms/connect-via
+     urls-stream
+     (fn [url]
+       (md/chain
+        (get-page! client-id client-secret url)
+        (fn [response]
+          (if (cpc/page-response-ok? response)
+            (let [resource (resource-key response)
+                  pagination (:pagination response)
+                  next-url (:next pagination)]
+              (if (blank? next-url)
+                (do (info "no more urls to fetch")
+                    (ms/close! urls-stream))
+                (ms/put! urls-stream next-url))
+              (ms/put-all! resources-stream resource))
+            (do (error "Error getting scans for url: " url)
+                (ms/close! urls-stream)
+                (Exception. "Error fetching scans."))))))
+     resources-stream)
+    resources-stream))
+
 (defn scans!
   "Returns a stream of historical scan results matching opts."
   [client-id client-secret opts]
-  (let [urls-stream (ms/stream 10) ;; absolutely no science here
-        scans-stream (ms/stream 20) ;; nor here
-        shovel (fn [url]
-                 (md/chain
-                  (get-page! client-id client-secret url)
-                  ;; TODO: error handling here?
-                  (fn [response]
-                    (let [{:keys [pagination scans]} response
-                          next-url (:next pagination)]
-                      (if (or (= "" next-url) (nil? next-url))
-                        (do (info "no more urls to fetch")
-                            (ms/close! urls-stream)
-                            (ms/close! scans-stream))
-                        (ms/put! urls-stream next-url))
-                      (ms/put-all! scans-stream scans)))))]
-    (ms/put! urls-stream (scans-url opts))
-    (ms/consume-async shovel urls-stream)
-    scans-stream))
+  (stream-paginated-resources! client-id client-secret (scans-url opts) :scans))
 
 (defn scans-with-details!
   "Returns a stream of historical scan results with their details.
@@ -92,20 +103,16 @@
   details (iff the details are FIM). See CloudPassage API docs for
   more illustration."
   [client-id client-secret scans-stream]
-  (let [scans-with-details (ms/stream 10)
-        add-details
-        (fn [scan]
-          (md/chain
-           (get-page! client-id client-secret (:url scan))
-           (fn [response]
-             (ms/put! scans-with-details (assoc scan :scan (:scan response)))
-             (when (ms/drained? scans-stream)
-               (ms/close! scans-with-details)))))]
-    (ms/consume-async add-details scans-stream)
-    ;; TODO: Figure out a way to automatically close the stream this function
-    ;;       returns without using the lower-level on-drained callback.
-    (ms/on-drained scans-stream #(ms/close! scans-with-details))
-    scans-with-details))
+  (let [scans-with-details-stream (ms/stream 10)]
+    (ms/connect-via
+     scans-stream
+     (fn [scan]
+       (md/chain
+        (get-page! client-id client-secret (:url scan))
+        (fn [response]
+          (ms/put! scans-with-details-stream (assoc scan :scan (:scan response))))))
+     scans-with-details-stream)
+    scans-with-details-stream))
 
 (defn ^:private report-for-module!
   "Get recent report data for a certain client, and filter based on module."
--- a/test/cloudpassage_lib/scans_test.clj
+++ b/test/cloudpassage_lib/scans_test.clj
@@ -52,8 +52,8 @@
     1 "svm"    ;; software version management
     2 "ccm"))  ;; configuration change management
 
-(def ^:private fake-pages 10)
-(def ^:private scans-per-page 10)
+(def ^:private fake-pages 3)
+(def ^:private scans-per-page 5)
 (def ^:private details-query-url
   (#'scans/scans-url {"details" "true"}))
 
@@ -62,7 +62,7 @@
    queries."
   [page-num next-page]
   {:scans (for [index-in-page (range scans-per-page)
-                :let [scan-index (+ (* scans-per-page page-num)
+                :let [scan-index (+ (* scans-per-page (dec page-num))
                                     index-in-page)]]
             {:scan-id scan-index
              :module (index->module scan-index)
@@ -92,7 +92,7 @@
   (let [parsed-url (u/url url)
         query (:query parsed-url)
         page-num (-> query
-                     (get "next" "0")
+                     (get "next" "1")
                      Integer/parseInt)
         next-page (if (< page-num fake-pages)
                     (str (assoc-in parsed-url [:query "next"] (inc page-num)))
@@ -112,16 +112,32 @@
    cskc/->snake_case_keyword
    (fake-get-page! client-id client-secret url)))
 
+(defn ^:private fake-get-page-with-bad-response!
+  "Like fake-get-page, but returns a bad status code."
+  [client-id client-secret url]
+  (is (= client-id "lvh"))
+  (is (= client-secret "hunter2"))
+  :cloudpassage-lib.core/fetch-error)
+
 (deftest scans!-tests
-  (with-redefs [scans/get-page! fake-get-page!]
-    (let [scans-stream (scans/scans! "lvh" "hunter2" {"modules" "fim"})
-          scans (ms/stream->seq scans-stream)]
-      (is (= (for [scan-id (range (* fake-pages scans-per-page))]
-               {:scan-id scan-id
-                :module (index->module scan-id)
-                :url details-query-url})
-             scans))
-      (is (ms/closed? scans-stream)))))
+  (testing "Successful scan with pagination."
+    (with-redefs [scans/get-page! fake-get-page!]
+      (let [scans-stream (scans/scans! "lvh" "hunter2" {"modules" "fim"})
+            scans (ms/stream->seq scans-stream)]
+        (is (= (for [scan-id (range (* fake-pages scans-per-page))]
+                 {:scan-id scan-id
+                  :module (index->module scan-id)
+                  :url details-query-url})
+               scans))
+        (is (ms/closed? scans-stream)))))
+  (testing "If an error occurs an empty result is returned."
+    ;;TODO: Replacing the error thing to detect it was logged might be helpful.
+    (with-redefs [scans/get-page! fake-get-page-with-bad-response!]
+      (let [scans-stream (scans/scans! "lvh" "hunter2" {"modules" "fim"})
+            result (clojure.string/join "" (ms/stream->seq scans-stream))]
+        (is (ms/drained? scans-stream))
+        (is (ms/closed? scans-stream))
+        (is (= result ""))))))
 
 (deftest scans-with-details!-tests
   (testing "Typical scan returns expected page details."
--- a/src/cloudpassage_lib/core.clj
+++ b/src/cloudpassage_lib/core.clj
@@ -84,6 +84,11 @@
         (error "error fetching events page:" (.getMessage exc))
         ::fetch-error)))))
 
+(defn page-response-ok?
+  "Returns false if there was an error retrieving a page."
+  [response]
+  (not= ::fetch-error response))
+
 (defn fetch-token!
   "Fetch an access token for the cloudpassage api that belongs to the client-id/secret pair.
    If a token doesn't exist in Redis, then this will hit the cloudpassage api to obtain one.


src/cloudpassage_lib/scans.clj
```diff
@@ -61,27 +62,49 @@
   (let [token (cpc/fetch-token! client-id client-secret (:fernet-key env))]
     (cpc/get-single-events-page! token url)))
 
+(defn ^:private page-response-ok?
+  [response]
+  (not= :cloudpassage-lib.core/fetch-error response))
```
ehashman: I wonder if it might make more sense to make this a public interface inside of cloudpassage-lib.core?

reaperhulk: If my primitive reading of this is correct, this is a private function that returns a boolean to tell whether a response from cloudpassage-lib.core was okay? If so, I agree that it should live in cloudpassage-lib.core (I do like how it looks in the code below so :+1: for the idea Tim. I have missed working in a language where I can use question marks, heh)

src/cloudpassage_lib/scans.clj
```diff
@@ -61,27 +62,49 @@
...
+
+(defn ^:private map-stream
+  "Maps an input stream to an output stream with some function.
+
+  The function is expected to accept the output stream and return another
+  function that is called repeatedly for each bit of input."
+  [input f]
+  (let [output (ms/stream)]
```
ehashman: According to the manifold docs, this stream has no buffer. It appears to be replacing streams with buffer sizes of 10/20. Is that what we want?

src/cloudpassage_lib/scans.clj
TimSimpsonR: Great point. Done.

src/cloudpassage_lib/scans.clj
TimSimpsonR: I don't know. I was looking through the docs and saw the code often uses the zero arg arity and figured I should, as 10 seemed arbitrary. My guess was that it created a buffer with whatever felt like a decent buffer size to the author. But I am basing that on nadda so I can add the explicit amount back if I need to.

ehashman: That might be more of a question for @lvh or @derwolfe, who would know better.

ehashman: I also wonder if it might make things a bit easier to understand if we passed the `output` as an argument to this function instead of instantiating the stream and binding it locally. That way, we wouldn't need to pass the function `f` to `map-stream` in the form `(fn [output] (fn [some-other-args...] ...))` and instead could just pass `(fn [some-other-args] ...)` and the output stream as an additional parameter.

It would also allow us to configure the buffer size of the output stream, if that's a feature.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -147,6 +163,21 @@
            ms/stream->seq
            doall))))
 
+(defn ^:private make-fake-page
```
ehashman: Maybe I'm missing something, but this function doesn't appear to be used anywhere?

src/cloudpassage_lib/scans.clj
```diff
@@ -61,27 +62,45 @@
...
     (cpc/get-single-events-page! token url)))
 
+(defn ^:private map-stream
+  "Maps an input stream to an output stream with some function.
+
+  The function is expected to accept the output stream and return another
+  function that is called repeatedly for each bit of input."
+  [input f]
+  (let [output (ms/stream)]
+    (ms/connect-via input (f output) output)
```
derwolfe: It seems like this might be the incorrect syntax - `(f output)` is invoking the function, it should probably be a callable like `(ms/connect-via input-stream #(f output-stream) output-stream)`

src/cloudpassage_lib/scans.clj
```diff
@@ -61,27 +62,45 @@
...
+
+(defn ^:private stream-paginated-resources!
+  "Returns a stream of resources coming from a paginated list."
+  [client-id client-secret initial-url resource-key]
+  (let [urls-stream (ms/stream 10)]
+    (ms/put! urls-stream initial-url)
+    (map-stream
+     urls-stream
+     (fn [scans-stream]
+       (fn [url]
```
derwolfe: I'm a bit curious if `map-stream` is an abstraction that makes this more or less understandable.

As a thought experiment, what if this were rewritten just using connect via. The implementation might look like:

```clojure
(defn ^:private stream-paginated-resources!
  "Returns a stream of resources coming from a paginated list."
  [client-id client-secret initial-url resource-key]
  (let [urls-stream (ms/stream 10)
        output-stream (ms/stream 10)
        ;; fetch-page, were it declared outside of the function
        ;; would need to take the output stream it wants to close
        ;; as an input. However, since it is locally bound, it is
        ;; free to be closed inside of stream-paginated resources.
        fetch-page (fn [url]
                       (md/chain
                        (get-page! client-id client-secret url)
                        (fn [response]
                          (if (cpc/page-response-ok? response)
                            (let [resource (resource-key response)
                                  pagination (:pagination response)
                                  next-url (:next pagination)]
                              (if (blank? next-url)
                                (do (info "no more urls to fetch")
                                    (ms/close! urls-stream))
                                (ms/put! urls-stream next-url))
                              (ms/put-all! output-stream resource))
                            (do (error "Error fetching url: " url)
                                (ms/close! urls-stream)
                                (Exception. "Error fetching scans."))))))]
    (ms/put! urls-stream initial-url)
    ;; invoke fetch-page each time a new url is pushed into the urls stream
    (ms/connect-via #(fetch-page %) urls-stream)
    output-stream ;;maybe there is a better name for this?
    ))
 ```
What do you think?

File: src/cloudpassage_lib/scans.clj (hunk 2 above)
Comment (TimSimpsonR) [inline src/cloudpassage_lib/scans.clj]: My motivation for building this was to make the output buffer get built by the function. That way the anything using the map-stream function doesn't need to worry about the managing the lifetime of the output buffer but can just use it. If there was a way to make the function passed in not need to know about the output stream at all but just return elements somehow I'd have done it.

File: test/cloudpassage_lib/scans_test.clj (hunk 3 above)
Comment (TimSimpsonR) [inline test/cloudpassage_lib/scans_test.clj]: Thanks- I forgot to delete it when I whipped this up.

File: src/cloudpassage_lib/scans.clj (hunk 2 above)
Comment (derwolfe) [inline src/cloudpassage_lib/scans.clj]: A comment I left earlier (https://github.com/RackSec/cloudpassage-lib/pull/33#discussion_r56868204) is somewhat related to this.

File: src/cloudpassage_lib/scans.clj (hunk 4 above)
Comment (TimSimpsonR) [inline src/cloudpassage_lib/scans.clj]: connect-via expects a function which it will passes data from input into repeatedly. So if it gets a function that returns a function it will fail as the arity will be wrong.

File: src/cloudpassage_lib/scans.clj (hunk 5 above)
Comment (TimSimpsonR) [inline src/cloudpassage_lib/scans.clj]: With map-stream, stream-paginated-resources is 23 lines, without it is 29, so in that sense I can see how the change seems superfluous. However using another function makes it obvious that the same code being used for both functions using map-stream (in my other PR with the server API calls, it is used in four different functions).

The bigger deal is that the lifetime of output stream is not being managed directly when map-stream is used. In general it seems like we don't want to be creating and closing streams in functions where most of the logic concerns a different subject like pagination details.

Comment (derwolfe) [inline src/cloudpassage_lib/scans.clj]: It could just be that we need to refactor what map-stream looks like, since it shouldn't need to take a nested function in order to process inputs; e.g., `(fn [stream] (fn [thing-from-stream] ...))` shouldn't be necessary in order to work.

File: src/cloudpassage_lib/scans.clj (hunk 4 above)
Comment (derwolfe) [inline src/cloudpassage_lib/scans.clj]: Tim - I'm not sure I follow, when I've used `connect-via`, I've passed it a function to execute each time the stream produces; e.g. https://github.com/RackSec/cloudpassage-poller/blob/master/src/cloudpassage_poller/core.clj#L169

Comment (derwolfe) [inline src/cloudpassage_lib/scans.clj]: Here are also some tests demonstrating how it works: https://github.com/RackSec/cloudpassage-poller/blob/master/test/cloudpassage_poller/core_test.clj#L52

File: src/cloudpassage_lib/scans.clj (hunk 5 above)
Comment (TimSimpsonR) [inline src/cloudpassage_lib/scans.clj]: Oh I totally missed the fact that connect-via makes things substantially shorter and we don't need to call on-drained anymore. I did that after I'd already moved it to a function.

I still like having a function manage the lifetime of the stream, but it is way less stuff than it was for using connect-via. I still like the idea of a function though because if we find a better solution than connect via that actually does return a stream or has other properties we like it will be easier to switch. Also there will still be four places ultimately where we'll end up using this.  I don't feel quite as strong about it now but I still believe giving it its own function is best.

Comment (TimSimpsonR) [inline src/cloudpassage_lib/scans.clj]: Agh, I didn't refresh and see your last message before I commented.

I only made it use nested functions because the function still has to write to the output stream. So it's really (fn [stream] ( fn [thing-from-different-stream] ...)) since the thing-from-stream is from the input stream. Does connect via accept a function that could return stream data?

File: src/cloudpassage_lib/scans.clj (hunk 4 above)
Comment (TimSimpsonR) [inline src/cloudpassage_lib/scans.clj]: It's working the same way here. The lambda which calls fetch-events! accepts a single parameter as does the function which is returned  by (f output).

File: src/cloudpassage_lib/scans.clj (hunk 5 above)
Comment (derwolfe) [inline src/cloudpassage_lib/scans.clj]: I'd vote for just using `connect-via`. My reasons being:

(a) I find it clearer. It has single responsibility - establishing a connection between two streams. Using connect-via explicitly communicates that if something closes upstream from the consumer, the consumer will finish consuming, then close itself.
(b) It requires no new test coverage. Currently, `map-stream` is only exercised by the nature of it being used by other functions; it doesn't however have tests for its specific behavior.

**Comment by TimSimpsonR** [comment]
@derwolfe map-stream taken out.

File: src/cloudpassage_lib/scans.clj [hunk 6]
```diff
@@ -61,27 +62,37 @@
   (let [token (cpc/fetch-token! client-id client-secret (:fernet-key env))]
     (cpc/get-single-events-page! token url)))
 
+(defn ^:private stream-paginated-resources!
+  "Returns a stream of resources coming from a paginated list."
+  [client-id client-secret initial-url resource-key]
+  (let [urls-stream (ms/stream 10)
+        output (ms/stream 20)]
+    (ms/put! urls-stream initial-url)
+    (ms/connect-via
```
derwolfe: I really like the use of `connect-via` here :-). Thanks!

src/cloudpassage_lib/scans.clj
```diff
@@ -61,27 +62,37 @@
   (let [token (cpc/fetch-token! client-id client-secret (:fernet-key env))]
     (cpc/get-single-events-page! token url)))
 
+(defn ^:private stream-paginated-resources!
+  "Returns a stream of resources coming from a paginated list."
+  [client-id client-secret initial-url resource-key]
+  (let [urls-stream (ms/stream 10)
+        output (ms/stream 20)]
```
derwolfe: Would you mind renaming output to something like `output-stream` or `page-stream` (or something else, just a bit more descriptive)?

TimSimpsonR: Done.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -94,7 +94,7 @@
         page-num (-> query
                      (get "next" "0")
                      Integer/parseInt)
-        next-page (if (< page-num fake-pages)
+        next-page (if (< page-num (- fake-pages 1))
```
ehashman: Might be an idea to make this a `<=` instead of decrementing the number of fake-pages.

TimSimpsonR: I made that same mistake for some reason, but the goal is to make the next-page stop appearing one earlier than it currently is, not one later. Which makes sense- the fake page you always get back should be the number of pages that had a next link plus one, so the number of pages with a next link is the expected fake count minus one.

ehashman:
Echoing Chris' suggestion of renaming `output-stream` to something more descriptive, I don't see the updated fix? Calling [this one](https://github.com/RackSec/cloudpassage-lib/pull/33/files#diff-8844961b8f6dc4756b96fd734c1fbc77R69) `page-stream` and [this one](https://github.com/RackSec/cloudpassage-lib/pull/33/files#diff-8844961b8f6dc4756b96fd734c1fbc77R106) `scans-with-details` will significantly improve readability imo.

Other than those comments, LGTM.

test/cloudpassage_lib/scans_test.clj
ehashman: Hm, that doesn't make sense to me though in the context of the bug. If we weren't fetching enough pages before due to the off by one bug, shouldn't we be looking at one additional page as opposed to one less page?

TimSimpsonR: You're right, however unless I'm missing a detail this function is what is used to redefine the get-page! call, but the expected results of the tests are generated with for statements. So what I observed with get-servers was that an extra page got returned and due to that the tests failed because of the extra results.

TimSimpsonR: To illustrate, if I take out the change I see this:
```clojure
FAIL in (scans!-tests) (scans_test.clj:127)
Successful scan with pagination.
expected: ({:scan-id 0,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 1,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 2,
            :module "ccm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 3,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 4,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 5,
            :module "ccm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 6,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 7,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 8,
            :module "ccm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 9,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 10,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 11,
            :module "ccm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 12,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 13,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 14,
            :module "ccm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="})
  actual: ({:scan-id 0,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 1,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 2,
            :module "ccm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 3,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 4,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 5,
            :module "ccm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 6,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 7,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 8,
            :module "ccm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 9,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 10,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 11,
            :module "ccm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 12,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 13,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 14,
            :module "ccm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 15,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 16,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 17,
            :module "ccm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 18,
            :module "fim",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="}
           {:scan-id 19,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules="})
    diff: + [nil
             nil
             nil
             nil
             nil
             nil
             nil
             nil
             nil
             nil
             nil
             nil
             nil
             nil
             nil
             {:scan-id 15,
              :module "fim",
              :url
              "https://api.cloudpassage.com/v1/scans?details=true&modules="}
             {:scan-id 16,
              :module "svm",
              :url
              "https://api.cloudpassage.com/v1/scans?details=true&modules="}
             {:scan-id 17,
              :module "ccm",
              :url
              "https://api.cloudpassage.com/v1/scans?details=true&modules="}
             {:scan-id 18,
              :module "fim",
              :url
              "https://api.cloudpassage.com/v1/scans?details=true&modules="}
             {:scan-id 19,
              :module "svm",
              :url
              "https://api.cloudpassage.com/v1/scans?details=true&modules="}]
```

ehashman: Aha, I figured out my issue. Here's a sample run of the API in action:

```
16-03-21 21:06:32 rackbox INFO [cloudpassage-lib.core] - fetching https://api.cloudpassage.com/v1/scans?modules=&since=2016-03-21T18%3A06%3A32Z
16-03-21 21:06:33 rackbox INFO [cloudpassage-lib.core] - fetching https://api.cloudpassage.com/v1/scans?page=2&per_page=10&since=2016-03-21T18%3A06%3A32Z
16-03-21 21:06:34 rackbox INFO [cloudpassage-lib.core] - fetching https://api.cloudpassage.com/v1/scans?page=3&per_page=10&since=2016-03-21T18%3A06%3A32Z
16-03-21 21:06:34 rackbox INFO [cloudpassage-lib.core] - fetching https://api.cloudpassage.com/v1/scans?page=4&per_page=10&since=2016-03-21T18%3A06%3A32Z
16-03-21 21:06:35 rackbox INFO [cloudpassage-lib.scans] - no more urls to fetch
```

As you can see there, pages are 1-indexed.

However, in our code, we are 0-indexing pages, per [L95](https://github.com/RackSec/cloudpassage-lib/pull/33/files#diff-5eb370bbc8927820e4beaf9400e871d4R95). I think we should make our fake paging match the API, and set that initial page number to 1 and not 0. That way, we don't need to decrement this here.

TimSimpsonR:
@ehashman Calling the first one "page-stream" would be incorrect, as it returns a stream of resources. For the second, are you saying to give it the same name as the function?

Also I found this code very difficult to reason about when I first went through it in large part because differentiating between which stream was input and which was output got tricky. Also not to beat a dead horse, but map-stream, reduced the places the output stream appeared which cut down the noise somewhat.

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: That's a great idea. Let me change it.

TimSimpsonR: Actually, let me take that back. The expected results are mocked independently enough that I'd need to change all of that code too, along with the code I just added in my other PR, and it would still just be dummy data anyway. I'd rather not.

ehashman:
`page-stream` was just the name Chris suggested. The function is called `stream-paginated-resources!`, so maybe you'd prefer `resource-stream`?

The issue with the name `output-stream` is that it doesn't tell me anything about what's *in* the output stream, which I find confusing. In `scans!`, it used to be called `scans-stream`, so I knew I was putting scans in there when reading through the code. Similarly, I know I'm putting scans with details into the `scans-with-details` stream, and I know that I am consuming URLs from the `urls-stream`. Would nice for this to be both illustrative and consistent.

TimSimpsonR:
@ehashman I agree that output-stream may be more informative than just output. "url-streams" is also a great name because there are no other clues that urls would be put there, other than of course the one line of code that puts the url there.

I feel putting the word "scans" into the function "scans-with-details" is redundant as the name of the function tells you what is being passed back. Ditto with stream-paginated-resources!

test/cloudpassage_lib/scans_test.clj
ehashman: Huh? All you need to do is change the "0" in L95 to "1", and ensure you [decrement `page-num` here](https://github.com/RackSec/cloudpassage-lib/pull/33/files#diff-5eb370bbc8927820e4beaf9400e871d4R65) in this scan-id generator to make the scan-id math work out.

TimSimpsonR: Not sure if the link to the "scan-id generator" is working for me as it takes me to the place where I'm decrementing the next-page for this PR.

Anyway here's what I see if I change "0" to "1":
```clojure

FAIL in (svm-report!-tests) (scans_test.clj:170)
expected: ({:scan-id 1,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules=",
            :scan {}}
           {:scan-id 4,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules=",
            :scan {}}
           {:scan-id 7,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules=",
            :scan {}}
           {:scan-id 10,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules=",
            :scan {}}
           {:scan-id 13,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules=",
            :scan {}})
  actual: ({:scan-id 7,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules=",
            :scan {}}
           {:scan-id 10,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules=",
            :scan {}}
           {:scan-id 13,
            :module "svm",
            :url
            "https://api.cloudpassage.com/v1/scans?details=true&modules=",
            :scan {}})
    diff: - [{:scan-id 1}
             {:scan-id 4}
             {:scan-id 7}
             {:scan-id 10,
              :module "svm",
              :url
              "https://api.cloudpassage.com/v1/scans?details=true&modules=",
              :scan {}}
             {:scan-id 13,
              :module "svm",
              :url
              "https://api.cloudpassage.com/v1/scans?details=true&modules=",
              :scan {}}]
          + [{:scan-id 7} {:scan-id 10} {:scan-id 13}]
16-03-21 21:41:11 MJV0HLDKQ4 INFO [cloudpassage-lib.scans] - no more urls to fetch
```

ehashman: I don't know why GH is being temperamental about that link. Here's the diff of the fix I tested:

```diff
diff --git a/test/cloudpassage_lib/scans_test.clj b/test/cloudpassage_lib/scans_test.clj
index 058f2fd..9cae320 100644
--- a/test/cloudpassage_lib/scans_test.clj
+++ b/test/cloudpassage_lib/scans_test.clj
@@ -62,7 +62,7 @@
    queries."
   [page-num next-page]
   {:scans (for [index-in-page (range scans-per-page)
-                :let [scan-index (+ (* scans-per-page page-num)
+                :let [scan-index (+ (* scans-per-page (dec page-num))
                                     index-in-page)]]
             {:scan-id scan-index
              :module (index->module scan-index)
@@ -92,9 +92,9 @@
   (let [parsed-url (u/url url)
         query (:query parsed-url)
         page-num (-> query
-                     (get "next" "0")
+                     (get "next" "1")
                      Integer/parseInt)
-        next-page (if (< page-num (- fake-pages 1))
+        next-page (if (< page-num fake-pages)
                     (str (assoc-in parsed-url [:query "next"] (inc page-num)))
                     "")]
     (when-some [since (query "since")]
```

PR #35, opened 2016-03-24
Fetch report data for each server

derwolfe:
Builds on #30.

Fetch reporting information for a set of servers instead of all reports for a time interval.

### DIFF
--- a/src/cloudpassage_lib/scans.clj
+++ b/src/cloudpassage_lib/scans.clj
@@ -20,6 +20,12 @@
 (def ^:private base-scans-url
   "https://api.cloudpassage.com/v1/scans/")
 
+(def ^:private scans-path
+  (partial str "/v1/scans/"))
+
+(def ^:private base-servers-url
+  "https://api.cloudpassage.com/v1/servers/")
+
 (defn ^:private maybe-flatten-list
   [maybe-list]
   (if (or (string? maybe-list) (nil? maybe-list))
@@ -35,7 +41,30 @@
          (update :query merge opts)
          str))))
 
-(defn get-page!
+(defn ^:private scan-server-url
+  "URL for fetching most recent scan results of a server."
+  [server-id module]
+  (str (u/url base-servers-url server-id module)))
+
+(defn ^:private scans-detail-url
+  "Compute the URL for a scan detail."
+  ([scan-id]
+   (scans-detail-url base-scans-url scan-id))
+  ([url scan-id]
+   (-> (u/url url)
+       (assoc :path (scans-path scan-id))
+       str)))
+
+(defn ^:private finding-detail-url
+  "Compute the URL for a particular finding in a particular scan."
+  ([scan-id finding-id]
+   (finding-detail-url base-scans-url scan-id finding-id))
+  ([url scan-id finding-id]
+   (-> (u/url url)
+       (assoc :path (scans-path scan-id "/findings/" finding-id))
+       str)))
+
+(defn ^:private get-page!
   "Gets a page, and handles auth for you."
   [client-id client-secret url]
   (let [token (cpc/fetch-token! client-id client-secret (:fernet-key env))]
@@ -62,11 +91,16 @@
                     (ms/close! urls-stream))
                 (ms/put! urls-stream next-url))
               (ms/put-all! resources-stream resource))
-            (do (error "Error getting scans for url: " url)
+            (do (error "Error getting scans for url:" url)
                 (ms/close! urls-stream)
                 (Exception. "Error fetching scans."))))))
      resources-stream)
     resources-stream))
+
+(defn list-servers!
+  "Returns a stream of servers for the given account."
+  [client-id client-secret]
+  (stream-paginated-resources! client-id client-secret base-servers-url :servers))
 
 (defn scans!
   "Returns a stream of historical scan results matching opts."
@@ -93,17 +127,37 @@
      scans-with-details-stream)
     scans-with-details-stream))
 
+(defn scan-each-server!
+  "Fetches a new report for each server passed in via the servers-stream.
+  The module determines which type of report will be fetched.
+
+  Returns a stream that will contain the a complete report for a given server."
+  [client-id client-secret module servers-stream]
+  (let [server-details-stream (ms/stream 10)
+        scan-server! (fn [server-id module]
+                       (let [url (scan-server-url server-id module)]
+                         (get-page! client-id client-secret url)))]
+    (ms/connect-via
+     servers-stream
+     (fn [{:keys [id]}]
+       (md/chain
+        (scan-server! id module)
+        (fn [response]
+          (if (cpc/page-response-ok? response)
+            (ms/put! server-details-stream response)
+            (error "Error getting scans for server" id)))))
+     server-details-stream)
+    server-details-stream))
+
 (defn ^:private report-for-module!
   "Get recent report data for a certain client, and filter based on module."
   [client-id client-secret module-name]
   ;; The docs say we can use "module" as a query parameter but it does
   ;; not work for FIM or SVM, so we have to filter out those items instead.
-  (let [opts {"since" (cpc/->cp-date (-> 3 hours ago))}]
-    (->> (scans! client-id client-secret opts)
-         (ms/filter (fn [{:keys [module]}] (= module module-name)))
-         (scans-with-details! client-id client-secret)
-         (ms/map #(cske/transform-keys cskc/->kebab-case-keyword %))
-         ms/stream->seq)))
+  (->> (list-servers! client-id client-secret)
+       (scan-each-server! client-id client-secret module-name)
+       (ms/map #(cske/transform-keys cskc/->kebab-case-keyword %))
+       ms/stream->seq))
 
 (defn fim-report!
   "Get the current (recent) FIM report for a particular client."
--- a/test/cloudpassage_lib/scans_test.clj
+++ b/test/cloudpassage_lib/scans_test.clj
@@ -4,12 +4,11 @@
    [cloudpassage-lib.core :refer [cp-date-formatter]]
    [clj-time.format :as tf]
    [clj-time.core :as t :refer [millis hours ago within?]]
+   [clojure.string :as str]
    [clojure.test :refer [deftest testing is are]]
    [manifold.stream :as ms]
    [manifold.deferred :as md]
    [cemerick.url :as u]
-   [camel-snake-kebab.core :as cskc]
-   [camel-snake-kebab.extras :as cske]
    [taoensso.timbre :as timbre :refer [info spy]]))
 
 (deftest scans-url-tests
@@ -31,6 +30,22 @@
       {"modules" "fim"}
       {"modules" "fim"
        "since" "2016-01-01"})))
+
+(deftest scans-detail-url-tests
+  (is (= "https://api.cloudpassage.com/v1/scans/abcdef"
+         (#'scans/scans-detail-url "abcdef")))
+  (is (= "https://abc.com/v1/scans/abcdef"
+         (#'scans/scans-detail-url "https://abc.com/" "abcdef"))))
+
+(deftest finding-detail-url-tests
+  (is (= "https://api.cloudpassage.com/v1/scans/abcdef/findings/xyzzy"
+         (#'scans/finding-detail-url "abcdef" "xyzzy")))
+  (is (= "https://abc.com/v1/scans/abcdef/findings/xyzzy"
+         (#'scans/finding-detail-url "https://abc.com/" "abcdef" "xyzzy"))))
+
+(deftest scan-server-url-tests
+  (is (= "https://api.cloudpassage.com/v1/servers/server-id/svm"
+         (#'scans/scan-server-url "server-id" "svm"))))
 
 (defn ^:private index->module
   "Given an index of a (fake, test-only) scan, return a module for that scan."
@@ -93,19 +108,22 @@
       (fake-details-page)
       (fake-scans-page page-num next-page))))
 
-(defn ^:private fake-get-page-with-snakes
-  "Return snake-cased versions of what is returned by `fake-get-page`"
-  [client-id client-secret url]
-  (cske/transform-keys
-   cskc/->snake_case_keyword
-   (fake-get-page! client-id client-secret url)))
-
-(defn ^:private fake-get-page-with-bad-response!
-  "Like fake-get-page, but returns a bad status code."
-  [client-id client-secret url]
-  (is (= client-id "lvh"))
-  (is (= client-secret "hunter2"))
-  :cloudpassage-lib.core/fetch-error)
+(defn ^:private use-atom-log-appender!
+  []
+  (let [log (atom [])
+        log-appender-fn (fn [data]
+                          (let [{:keys [output-fn]} data
+                                formatted-output-str (output-fn data)]
+                            (swap! log conj formatted-output-str)))]
+    (timbre/merge-config!
+     {:appenders
+      {:atom-appender
+       {:async false
+        :enabled? true
+        :min-level nil
+        :output-fn :inherit
+        :fn log-appender-fn}}})
+    log))
 
 (deftest scans!-tests
   (testing "Successful scan with pagination."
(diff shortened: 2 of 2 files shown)


src/cloudpassage_lib/scans.clj
```diff
@@ -56,7 +64,7 @@
        (assoc :path (scans-path scan-id "/findings/" finding-id))
        str)))
 
-(defn get-page!
+(defn ^:private get-page!
```
ehashman: +1

test/cloudpassage_lib/scans_test.clj
```diff
@@ -112,12 +117,15 @@
...
 
-(defn ^:private fake-get-page-with-bad-response!
-  "Like fake-get-page, but returns a bad status code."
-  [client-id client-secret url]
-  (is (= client-id "lvh"))
-  (is (= client-secret "hunter2"))
-  :cloudpassage-lib.core/fetch-error)
+(defn ^:private use-atom-log-appender!
+  [log log-appender-fn]
+  (timbre/merge-config! {:appenders
```
ehashman: The map arg here might look a bit better if you put it on the next line.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -130,22 +138,29 @@
...
+  (testing "When an error is encountered, the error is logged and an empty list is returned"
+    (let [log (atom [])
+          log-appender-fn (fn [data]
+                            (let [{:keys [output-fn]} data
+                                  formatted-output-str (output-fn data)]
+                              (swap! log conj formatted-output-str)))
+          expected-error (str "Error getting scans for url: "
+                              "https://api.cloudpassage.com/v1/scans?modules=fim")]
+      (with-redefs [scans/get-page! (fn [_client-id _client-secret _url]
+                                      :cloudpassage-lib.core/fetch-error)]
```
ehashman: It might be nicer to put this lambda in the let and give it a name instead of defining it in the with-redefs block.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -130,22 +139,29 @@
...
+        (use-atom-log-appender! log log-appender-fn)
+        (let [scans-stream (scans/scans! "lvh" "hunter2" {"modules" "fim"})
+              result (ms/stream->seq scans-stream)]
+          (is (str/includes? (first @log) expected-error))
```
TimSimpsonR: Very cool.

test/cloudpassage_lib/scans_test.clj
lvh: Yep. You may also just want to use `constantly` (at which point the definition becomes so short that it matters a bit less).

lvh: (To clarify; those are orthogonal points; you can put it in the let and still use `constantly`)

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,27 +179,114 @@
...
+      (let [server-stream (scans/list-servers! "lvh" "hunter2")
+            id-list (map :id (ms/stream->seq server-stream))]
+        (is (= ["server-id-1" "server-id-2" "server-id-3"] id-list))
+        (is (ms/closed? server-stream)))))
+  (testing "On error, fetching servers stops and the error is logged"
+    (let [log (atom [])
+          log-appender-fn (fn [data]
+                            (let [{:keys [output-fn]} data
+                                  formatted-output-str (output-fn data)]
+                              (swap! log conj formatted-output-str)))]
```
ehashman: This is the second time I've seen the log-appender-fn in a let binding I think, so we should probably factor that out.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,27 +179,114 @@
...
+    (with-redefs [scans/get-page!
+                  (fn [client-id client-secret url]
+                    (let [{:keys [page-num path]}
+                          (parse-fake-request client-id client-secret url)]
+                      (is (= "/v1/servers" path))
+                      (case page-num
+                        1 {:servers [{:id "server-id-1"} {:id "server-id-2"}]
+                           :pagination
+                           {:next (str @#'scans/base-servers-url "?page=2")}}
+                        2 {:servers [{:id "server-id-3"}]})))]
```
ehashman: In the next few tests, `scans/get-page!` is redef'd to a long inline anonymous function. While I think it's good to be clear about the functionality we're replacing `get-page!` with (so it might be obfuscating or confusing to wrap all the various functionality for these test fakes into `fake-get-page!`), this code appears to be copied verbatim below on line 214.

I would like to see a `fake-servers-page` in the style of `fake-scans-page`; we can chat about whether that should live inside `fake-get-page!` or not.

I'd also prefer to see this faked out in a more abstracted, configurable for loop, similar to how `fake-scans-page` works, as opposed to the hard-coded two pages.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,27 +179,114 @@
...
+      (with-redefs [scans/get-page!
+                    (fn [client-id client-secret url]
+                      (let [{:keys [page-num path]}
+                            (parse-fake-request client-id client-secret url)]
+                        (is (= "/v1/servers" path))
+                        (case page-num
+                          1 {:servers [{:id "server-id-1"} {:id "server-id-2"}]
+                             :pagination
+                             {:next (str @#'scans/base-servers-url "?page=2")}}
+                          2 :cloudpassage-lib.core/fetch-error)))]
```
ehashman: Oh, I guess this isn't totally copied verbatim; I missed case 2 being different. But I still think this can be factored out.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,27 +179,114 @@
...
+              id-list (map :id (ms/stream->seq server-stream))
+              expected-error (str "Error getting scans for url: "
+                                  "https://api.cloudpassage.com/v1/servers/?page=2")]
+          (is (str/includes? (first @log) expected-error)
+              "logs the page fetch error")
+          (is (= ["server-id-1" "server-id-2"] id-list))
+          (is (ms/closed? server-stream)))))))
+
+(deftest scan-each-server!-tests
+  (testing "Scan can handle an empty stream."
```
ehashman: This is a small nit/question: is an empty stream the same thing as a closed stream? (I don't think they are, but I also am not an expert at manifold.)

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,27 +179,114 @@
...
+    (let [input (ms/stream)]
+      (ms/close! input)
+      (let [scan-stream (scans/scan-each-server! "lvh" "hunter2" "svm" input)
+            scan-result (ms/stream->seq scan-stream)]
+        (is (empty? scan-result)))))
+  (testing "Returns an empty list when get-page returns ::fetch-error"
```
ehashman: Small typo: `:fetch-error` has two colons.

test/cloudpassage_lib/scans_test.clj
lvh: Nope. My understanding:

- An empty stream is a stream that currently has no items in it
- A closed stream is a stream that will not accept new items
- A drained stream is a stream that will not, and will never produce new items (i.e. empty + closed)

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,27 +179,114 @@
...
+    (with-redefs [scans/get-page!
+                  (fn [client-id client-secret url]
+                    (let [{:keys [path]}
+                          (parse-fake-request client-id client-secret url)]
+                      (is (= "/v1/servers/server-id-here/svm" path))
```
ehashman: I am not a fan of having this `is` here inside this redef'd function; I feel like `parse-fake-request` should be covered by tests elsewhere. Would it make this test worse if we used `(constantly :error-thing)` or similar instead?

test/cloudpassage_lib/scans_test.clj
lvh: Are you saying it should have 1? `::kw` is an automatically namespaced keyword; when you say `::kw` in the `cloudpassage-lib.core` namespace that automatically expands to `:cloudpassage-lib.core/fetch-error`. So if it should have 1 colon, it should have the namespace attached; otherwise it's a different keyword.

ehashman: Ohhh I didn't realize the double colon had a special meaning, I thought it was just a typo. Thanks!

derwolfe: I'll add the namespace prefix.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,27 +179,114 @@
...
+                  (fn [client-id client-secret url]
+                    (let [{:keys [path]}
+                          (parse-fake-request client-id client-secret url)]
+                      (case path
+                        "/v1/servers/server-1/svm" :one
+                        "/v1/servers/server-2/svm" :two
+                        "/v1/servers/server-3/svm" :three)))]
+      ;; note - if the input-stream's buffer is too small; then this test will
+      ;; fail. `put-all` must be able to place all elements in the stream''
+      ;; buffer size is set to 2, as it will be able to fill up with 3 elements
```
ehashman: I'm confused by two parts of this comment.

1. Wouldn't the buffer need to be size 3 to hold 3 things? Why is size 2 sufficient? This is just my lack of understanding of manifold.
2. Presumably this is a necessary condition because otherwise `put-all!` will block as nothing is consuming from the stream. Would it make more sense to test this in a more realistic scenario, i.e. calling `list-servers!` and consuming that with `scan-each-server!` in a threading macro? We're not actually doing that explicitly anywhere here, although I think we end up testing it implicitly through the report test.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -163,27 +179,114 @@
...
-    (let [report (report-fn! "lvh" "hunter2")]
-      (is (= (for [scan-id (range (* fake-pages scans-per-page))
-                   :let [module (index->module scan-id)]
-                   :when (= module expected-module)]
-               {:scan-id scan-id
-                :module module
-                :url details-query-url
-                :scan {}})
-             report))))
-  (with-redefs [scans/get-page! fake-get-page-with-snakes]
```
ehashman: `fake-get-page-with-snakes` got (accidentally?) removed here, and now I don't think it's getting used anywhere. We should still be testing this functionality imo.

test/cloudpassage_lib/scans_test.clj
derwolfe: The buffer of 0 means it can take a single item and hold it; a bucket. If you have a buffer size of 2, then the queue can hold 3 things.

test/cloudpassage_lib/scans_test.clj
derwolfe: While I agree that this *could* live in get-page, I don't believe that it *must*. Does the test perform the necessary function to be adequate to exercise the code? I believe it does.

ehashman:
Overall architectural comment on the tests: one of my concerns with these new tests is that each one redef's `get-page!` in a unique way. Previously, we faked all of `get-page!`'s functionality in one place, such that different tests calling the fake version would consistently get the same results given a particular URL, and thus we could ensure good integration testing. (By that I mean, we could redef `get-page!` to our single `fake-get-page!` function, and it would do the right thing if we strung together `scans!` and `scans-with-details!`, both of which call `get-page!`.)

By faking `get-page!` a different way each time, I worry that we are increasing our chances of introducing error into the tests, because we aren't requiring `get-page!` to behave consistently. Each one of these individual fakes introduces multiple different cases of behaviour; I guess I'm worried that the scope of the test behaviour itself is now large and complex enough to make me want to see *it* tested, because I found the size and scope of it difficult to reason about.

Ideally, what I would like to see is consistent behaviour for something that fakes the server listing API call and the server scan API call. Hence, I think we need to factor out all these individual redefs. Furthermore, we know that we want the server scan API call to be equivalent to a scan returned with details, so we might want to test for that too. While this doesn't all need to live behind/inside `fake-get-page!`, I do want all the list server API calls going to one function, all the server scan API calls going to one function, etc. as opposed to mocking out a bunch of individual cases between different tests.

Does that make sense?

test/cloudpassage_lib/scans_test.clj
ehashman: Thanks for the clarification :)

test/cloudpassage_lib/scans_test.clj
ehashman: Okay. Not gating or anything, but it would be nice if this string was be updated to be more accurate, since it looks like it's testing a closed, empty stream as opposed to just an empty stream. Unless I'm reading this test wrong.

test/cloudpassage_lib/scans_test.clj
TimSimpsonR: @ehashman I am not 100% clear on this either. @derwolfe and I discussed it for a bit because it was (and honestly still is) confusing for me, but here's my summary.

* When the calls are made to put-all and put since they are deferreds they end up getting executed in a thread somewhere. Same thing with close!. So when the code used to have three calls to put! instead of put-all!, it somehow worked; the three puts! would execute, and then the close!. My guess is its because the scheduler or whatever behind the scenes gave the puts! priority and made sure those worked before it executed the call to close!

* However put-all is all by itself in another thread, and it can only store (incr buffer-size) number of elements when you pass it a vector. As a C++ fan my first thought was it was about being wary of resource constraints but as @derwolfe pointed out it's really about back pressure, since you could have a vector with one enormous element that would get put onto a stream with a buffer size of 1 while put-all! with a vector of three integers would only put the first element on the stream.

Where I am still fuzy is on when a put-all! followed by a close! would ever be safe if you didn't know the buffer size, say if it was somehow abstracted away or you were calling put-all in a general function that accepted a stream and might, as far as you knew be called by close! later.

test/cloudpassage_lib/scans_test.clj
derwolfe: You are totally correct - I'll modify this so that it tests the snake->kebab casing logic.

test/cloudpassage_lib/scans_test.clj
derwolfe: I'm not a big fan of it either; but I don't think that we really need to have a test for `parse-fake-request`- it is a helper function that solely exists for the tests. It seems like this is the only place (that I can think of) that allows to assert which path was hit.

test/cloudpassage_lib/scans_test.clj
derwolfe: I agree that the two functions (the mocks) are very similar. Personally, I like that the behavior we are wanting to elicit from the mock is directly visible and obvious inside of the test. In my opinion, it makes it easier for me to read. Were it factored out, we would then have a configurable, more complicated mock, that wouldn't be as easy to follow.

test/cloudpassage_lib/scans_test.clj
ehashman: Wow, when I wrote that comment I just did a word substitution and didn't even mentally process that `parse-fake-request` is something we defined within the tests---I guess we are testing our tests here! Honestly, I don't think that's necessary, we can just return an error here imo. Will make it cleaner.

derwolfe: No problem. Just so I am clear, you'd prefer that I drop the path assertion, and just return the error?

ehashman: Yes. And since there's a few other places we do that, it might make sense to factor that out.

derwolfe: Sorry - What do you mean by factor it out?

ehashman: Ah, I mean write a helper along the lines of `return-error-page` or something like that? There used to be one before.

src/cloudpassage_lib/scans.clj
```diff
@@ -38,6 +41,11 @@
          (update :query merge opts)
          str))))
 
+(defn ^:private scan-server-url
+  "URL for fetching most recent scan results of a server."
+  [server-id module]
+  (str (u/url "https://api.cloudpassage.com/v1/servers/" server-id module)))
```
ehashman: Derp moment for me: this should be using `base-servers-url` instead of the hardcoded string right?

derwolfe: Nice find. I think so :-).

derwolfe:
@ehashman I agree that having the logic that represents that cloudpassage api located in a single place is useful from both the a maintenance and an understandability perspective. The issue I've come up against with the current mock is that it doesn't allow for any failure path behaviors, e.g. it cannot respond with a single page of servers or scans and then produce a `:fetch-error`. 

With that in mind, I don't feel like creating a single mock for the happy path behaviors would be that useful, as that mock wouldn't be used when we need to test for exceptional behavior.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -93,19 +108,22 @@
...
-   cskc/->snake_case_keyword
-   (fake-get-page! client-id client-secret url)))
-
-(defn ^:private fake-get-page-with-bad-response!
-  "Like fake-get-page, but returns a bad status code."
-  [client-id client-secret url]
-  (is (= client-id "lvh"))
-  (is (= client-secret "hunter2"))
-  :cloudpassage-lib.core/fetch-error)
+(defn ^:private use-atom-log-appender!
```
derwolfe: @ehashman - this is the factored out log function. It now takes constructs the atom on its own, contains the appender, and returns the log to use for assertions.

ehashman: Very nice!

PR #43, opened 2016-04-01
Retry logic for HTTP errors

### LINKED ISSUE #41: Add retry logic on HTTP errors
I got my first 502 gateway error from the CloudPassage SCA API today! If we end up losing a single report like that, for our purposes the combined report is incomplete and hence incorrect, so we need to start over from scratch.

But it's wasteful to have to throw all the fetched data away in order to try again. Ideally, we should add some retry logic on encountering an error, and only fail after trying some number of times.

ehashman:
Fixes #41.

WIP. Next steps:

- [x] Figure out what to do instead of passing the symbol "fail", probably throw an exception
- [x] Update `get-page!` to wrap `get-page-retry!` 
- [x] Remove the error handling inside various functions inside scans

### DIFF
--- a/src/cloudpassage_lib/scans.clj
+++ b/src/cloudpassage_lib/scans.clj
@@ -7,6 +7,7 @@
    [aleph.http :as http]
    [manifold.deferred :as md]
    [manifold.stream :as ms]
+   [manifold.time :as mt]
    [environ.core :refer [env]]
    [cloudpassage-lib.core :as cpc]
    [taoensso.timbre :as timbre :refer [error info spy]]
@@ -56,7 +57,9 @@
            (throw (Exception. "Error fetching scans.")))
        :else
        (do (error "Couldn't fetch page. Retrying.")
-           (get-page-retry! token url (dec num-retries)))))))
+           (mt/in
+            3000
+            #(get-page-retry! token url (dec num-retries))))))))
 
 (defn ^:private get-page!
   "Gets a page, and handles auth for you."
--- a/src/cloudpassage_lib/scans.clj
+++ b/src/cloudpassage_lib/scans.clj
@@ -46,13 +46,17 @@
 (defn ^:private get-page-retry!
   "Gets a page, and handles auth for you."
   [token url num-tries]
-  (let [response (cpc/get-single-events-page! token url)]
-    (cond
-      (cpc/page-response-ok? response) response
-      (zero? num-tries) 'fail
-      :else
-      (do (error "Couldn't fetch page. Retrying.")
-          (recur token url (dec num-tries))))))
+  (md/chain
+   (cpc/get-single-events-page! token url)
+   (fn [response]
+     (cond
+       (cpc/page-response-ok? response) response
+       (zero? num-tries)
+       (do (error "No more retries.")
+           (throw (Exception. "Error fetching scans.")))
+       :else
+       (do (error "Couldn't fetch page. Retrying.")
+           (get-page-retry! token url (dec num-tries)))))))
 
 (defn ^:private get-page!
   "Gets a page, and handles auth for you."
--- a/src/cloudpassage_lib/scans.clj
+++ b/src/cloudpassage_lib/scans.clj
@@ -44,25 +44,25 @@
   (str (u/url base-servers-url server-id module)))
 
 (defn ^:private get-page-retry!
-  "Gets a page, and handles auth for you."
-  [token url num-tries]
+  "Gets a page, and handles retries on error."
+  [token url num-retries]
   (md/chain
    (cpc/get-single-events-page! token url)
    (fn [response]
      (cond
        (cpc/page-response-ok? response) response
-       (zero? num-tries)
+       (zero? num-retries)
        (do (error "No more retries.")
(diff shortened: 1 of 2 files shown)


src/cloudpassage_lib/scans.clj
```diff
@@ -43,6 +43,17 @@
   [server-id module]
   (str (u/url base-servers-url server-id module)))
 
+(defn ^:private get-page-retry!
+  "Gets a page, and handles auth for you."
+  [token url num-tries]
+  (let [response (cpc/get-single-events-page! token url)]
+    (cond
+      (cpc/page-response-ok? response) response
+      (zero? num-tries) 'fail
```
derwolfe: Interesting! What is `'fail`? It seems like it is intended to be used as a keyword, but isn't a keyword. :-)

src/cloudpassage_lib/scans.clj
```diff
@@ -43,6 +43,17 @@
   [server-id module]
   (str (u/url base-servers-url server-id module)))
 
+(defn ^:private get-page-retry!
+  "Gets a page, and handles auth for you."
+  [token url num-tries]
+  (let [response (cpc/get-single-events-page! token url)]
```
derwolfe: One issue that this might have is that `get-single-events-page!` returns a deferred. As such, this will not wait for the request to complete before iterating again. In order for this to work, the deferred returned by `get-single-events-page!` will need to be either unwrapped via a consuming function (see https://github.com/RackSec/cloudpassage-lib/blob/master/test/cloudpassage_lib/core_test.clj#L31 for an example) or dereferenced (using `@`).

test/cloudpassage_lib/scans_test.clj
```diff
@@ -35,6 +52,25 @@
   (is (= "https://api.cloudpassage.com/v1/servers/server-id/svm"
          (#'scans/scan-server-url "server-id" "svm"))))
 
+(deftest get-page-retry!-tests
+  (testing "Returns 'fail after three tries"
+    (let [fake-get (constantly :cloudpassage-lib.core/fetch-error)]
```
derwolfe: Nice tests! The only issue I can see is that the tests handle the synchronous case where a deferred isn't involved. In order for this to function with `get-single-events-page!` (or any other deferred-returning function), this will need to handle unwrapping the value from the deferred. Here are a few examples of how to handle errors and success conditions (https://github.com/RackSec/cloudpassage-lib/blob/master/test/cloudpassage_lib/core_test.clj#L31)

src/cloudpassage_lib/scans.clj
```diff
@@ -43,6 +43,21 @@
...
+   (cpc/get-single-events-page! token url)
+   (fn [response]
+     (cond
+       (cpc/page-response-ok? response) response
+       (zero? num-tries)
+       (do (error "No more retries.")
+           (throw (Exception. "Error fetching scans.")))
+       :else
+       (do (error "Couldn't fetch page. Retrying.")
+           (get-page-retry! token url (dec num-tries)))))))
```
ehashman: This is now an actual recursive call because recur binds to the lambda, which is not what we want. I'm not sure how problematic this is; since num-tries is going to be pretty small (3?), we're not really at risk of a stack overflow.

Unsure if there is a better way to do this.

derwolfe: To be honest, I'm not sure either. I feel like experimenting a bit :-).

test/cloudpassage_lib/scans_test.clj
```diff
@@ -35,6 +52,34 @@
   (is (= "https://api.cloudpassage.com/v1/servers/server-id/svm"
          (#'scans/scan-server-url "server-id" "svm"))))
 
+(deftest get-page-retry!-tests
```
derwolfe: These are awesome.

src/cloudpassage_lib/scans.clj
```diff
@@ -43,11 +43,26 @@
...
+   (cpc/get-single-events-page! token url)
+   (fn [response]
+     (cond
+       (cpc/page-response-ok? response) response
+       (zero? num-retries)
+       (do (error "No more retries.")
+           (throw (Exception. "Error fetching scans.")))
+       :else
+       (do (error "Couldn't fetch page. Retrying.")
+           (get-page-retry! token url (dec num-retries)))))))
```
derwolfe: My intuition is that `in` would wrap `get-page-retry...`.

src/cloudpassage_lib/scans.clj
```diff
@@ -43,11 +44,28 @@
...
+   (fn [response]
+     (cond
+       (cpc/page-response-ok? response) response
+       (zero? num-retries)
+       (do (error "No more retries.")
+           (throw (Exception. "Error fetching scans.")))
+       :else
+       (do (error "Couldn't fetch page. Retrying.")
+           (mt/in
+            3000
```
derwolfe: What would you think of making this a parameter?

ehashman: Oh yeah, that seems eminently reasonable.

derwolfe:
Nice work! This looks good to go :-)

PR #46, opened 2016-04-07
stream and parse the responses; fixes #45

### LINKED ISSUE #45: JSON parsing should stream from a reader
Cheshire can handle parsing of a stream; it shouldn't use the `bs/to-string api`. Instead it should use:

```clojure
...
byte-stream/to-reader
#(cheshire.core/parse-stream % true)
)
```

https://github.com/RackSec/alertlogic-differ/blob/b6bc20a24c9c6f20d085521029c2f2a49d3df21c/src/alertlogic_differ/api.clj#L25

derwolfe:


ehashman:
LGTM!

PR #47, opened 2016-04-08
move atom log appender into a utility ns

derwolfe:
While working on a different PR I noticed that this utility could be applicable to namespaces outside of `scans`.


derwolfe:
This is also strictly a move - the functionality provided by `use-atom-log-appender` is unaltered.

ehashman:
Yup, this looks good to me and means I can lift it from the lib for testing in the reporter. Very good.

PR #48, opened 2016-04-08
Auth error handling

derwolfe:
1. Adds exponential backoff/retry logic for deferred generating functions.
2. Changes `get-auth-token!` to return a deferred, instead of dereferencing the deferred and returning the token.
3. Wraps the `http/post` inside of `get-auth-token!` with the retry logic from (1)

### DIFF
--- a/src/cloudpassage_lib/core.clj
+++ b/src/cloudpassage_lib/core.clj
@@ -1,7 +1,6 @@
 (ns cloudpassage-lib.core
   (:require
    [clojure.string :as str]
-   [clojure.math.numeric-tower :as math]
    [aleph.http :as http]
    [environ.core :refer [env]]
    [manifold.deferred :as md]
@@ -13,6 +12,7 @@
    [clj-time.core :as time]
    [clj-time.format :as f]
    [cloudpassage-lib.fernet :as fernet]
+   [cloudpassage-lib.retry :as retry]
    [cheshire.core :as json]))
 
 ;; the url from which new auth-tokens can be obtained.
@@ -45,58 +45,6 @@
   [date]
   (f/unparse cp-date-formatter date))
 
-(defn exponentially
-  "Returns a function that when evaluated will produce the initial wait
-  raised to the number of failures.
-
-  This is intended to be used in conjunction with other 2-arg combinators,
-  e.g., `up-to`."
-  [wait]
-  (fn [failures]
-    (math/expt wait (count failures))))
-
-(defn up-to
-  "Returns a function that when evaluated will either
-
-  1) return a number to use to wait until retrying a function again. Or,
-  2) throw an exception because the maximum number of retries, `stop` has
-  been reached.
-
-  This is intended to be used in conjunction with other 2-arg combinators,
-  e.g., `up-to`."
-  [stop retry?]
-  (fn [failures]
-    (if (< (count failures) stop)
-      (retry? failures)
-      (throw (last failures)))))
-
-(defn retry
-  "Takes a function that returns a `manifold.deferred/deferred`. Retries that
-  function until it succeeds or the number of failures equal the stop value.
-
-  `retry` expects to encounter exceptions when retrying `f`. As such,
-  it will catch all exceptions that `f` might throw and continue retrying.
-
-  f - a function that should be retried; must return a
-      `manifold.deferred/deferred'
-  p - an int representing the initial number of seconds to wait before retrying.
-      This will grow exponentially for each attempt.
-  stop - an int representing the number of tries that the api should make before
-         giving up and returning `:cloudpassage-lib.core/retry-failure`.
-
-  Returns a deferred wrapping the results of `f`."
-  [f p stop]
-  (let [try-after (->> (exponentially p)
-                       (up-to stop))]
-    (md/loop [failures []]
-      (md/catch
-       (f)
-       Exception
-        (fn [exc]
-          (let [all-failures (conj failures exc)
-                wait (try-after all-failures)]
-            (mt/in (mt/seconds wait) #(md/recur all-failures))))))))
-
 (defn get-auth-token!
   "Using the secret key and an ID, fetch a new auth token.
 
@@ -112,7 +60,7 @@
         starting-retry 4
         stop-after 3]
     (md/chain
-     (retry
+     (retry/retry-exp-backoff
       #(http/post auth-uri {:headers auth-header})
       starting-retry
       stop-after)
--- a/src/cloudpassage_lib/core.clj
+++ b/src/cloudpassage_lib/core.clj
@@ -46,8 +46,8 @@
   (f/unparse cp-date-formatter date))
 
 (defn retry
-  "Takes a function that returns a deferred. Retries that function
-  until it succeeds or the number of failures equal the stop value
+  "Takes a function that returns a `manifold.deferred/deferred`. Retries that
+  function until it succeeds or the number of failures equal the stop value.
 
   `retry` expects to encounter exceptions when retrying `f`. As such,
   it will catch all exceptions that `f` might throw and continue retrying.
@@ -57,7 +57,7 @@
   f - a function that should be retried; must return a
       `manifold.deferred/deferred'
   stop - an int representing the number of tries that the api should make before
-         giving up and throwing an exception/dying.
+         giving up and returning `:cloudpassage-lib.core/retry-failure`.
 
   Returns a deferred wrapping the results of `f`."
   [p f stop]
--- a/src/cloudpassage_lib/core.clj
+++ b/src/cloudpassage_lib/core.clj
@@ -61,16 +61,24 @@
 
   Returns a deferred wrapping the results of `f`."
   [f p stop]
-  (md/loop [tries 1]
-    (md/catch
-     (f)
-     Exception
-      (fn [exc]
-        (error "Failure retrying:" exc)
-        (if (= tries stop)
-          (throw exc))
-        (let [wait (mt/seconds (math/expt p tries))]
-          (mt/in wait #(md/recur (inc tries))))))))
+  (let [;; try-after is essentially the back-off strategy
+        ;; it does not handle retrying itself, but is responsible
+        ;; for determining whether to continue retrying and
+        ;; if so, how long to wait until the next retry.
+        try-after (fn [exceptions]
+                    (let [ct (count exceptions)
+                          next-wait (math/expt p ct)]
+                      (if (< ct stop)
+                        next-wait
+                        (throw (last exceptions)))))]
+    (md/loop [failures []]
+      (md/catch
+       (f)
+       Exception
+        (fn [exc]
+          (let [all-failures (conj failures exc)
+                wait (try-after all-failures)]
+            (mt/in (mt/seconds wait) #(md/recur all-failures))))))))
 
 (defn get-auth-token!
   "Using the secret key and an ID, fetch a new auth token.
--- a/src/cloudpassage_lib/core.clj
+++ b/src/cloudpassage_lib/core.clj
@@ -106,10 +106,7 @@
       #(http/post auth-uri {:headers auth-header})
       starting-retry
       stop-after)
-     (fn [val]
-       (if (= ::retry-failure val)
-         ::auth-failure
-         (json/parse-stream (bs/to-reader (:body val)) true))))))
+     #(json/parse-stream (bs/to-reader (:body %)) true))))
 
 (defn get-single-events-page!
   "get a page at `uri` using the provided `auth-token`.
--- a/src/cloudpassage_lib/core.clj
+++ b/src/cloudpassage_lib/core.clj
@@ -54,13 +54,12 @@
   [client-id client-key]
   (info "fetching new auth token for" client-id)
   (let [sent-at (time/now)
-        auth-header (->basic-auth-header client-id client-key)
-        token @(md/chain
-                (http/post auth-uri {:headers auth-header})
-                :body
-                bs/to-reader
-                #(json/parse-stream % true))]
-    token))
+        auth-header (->basic-auth-header client-id client-key)]
+    (md/chain
+     (http/post auth-uri {:headers auth-header})
+     :body
+     bs/to-reader
+     #(json/parse-stream % true))))
 
 (defn get-single-events-page!
   "get a page at `uri` using the provided `auth-token`.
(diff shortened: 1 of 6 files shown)


src/cloudpassage_lib/core.clj
```diff
@@ -100,7 +146,8 @@
       ;; a token is in redis
       (fernet/decrypt fernet-key token)
       ;; no token is present, fetch a new one
-      (let [new-token (get-auth-token! client-id client-secret)
+      ;; the operation to fetch the token should be asynchronous
+      (let [new-token @(get-auth-token! client-id client-secret)
```
derwolfe: This needs to handle `::auth-failure`

src/cloudpassage_lib/core.clj
```diff
@@ -55,12 +96,17 @@
...
+        starting-retry 4
+        stop-after 3]
+    (md/chain
+     (retry
+      #(http/post auth-uri {:headers auth-header})
+      starting-retry
+      stop-after)
+     (fn [val]
+       (if (= ::retry-failure val)
+         ::auth-failure
```
derwolfe: Seems like returning something more like `[data, err]` would be a better approach. Making the return type of the function different is kind of gross.

src/cloudpassage_lib/core.clj
```diff
@@ -44,23 +45,69 @@
...
+  Returns a deferred wrapping the results of `f`."
+  [f p stop]
+  (let [invoker (fn []
+                  (->
+                   (f)
+                   (md/catch
+                    Exception
+                    (fn [exc]
+                      (error "Failure retrying:" (.getMessage exc))
+                      ::local-retry-error))))]
```
lvh: So we're using keywords to signal the failure now? I thought we were using the error deferreds?

src/cloudpassage_lib/core.clj
```diff
@@ -44,23 +45,69 @@
...
+           (and (= tries stop) (= ::local-retry-error val))
+           (let [d (md/deferred)]
+             (error "Failed retrying" tries "times; stopping")
+             (md/error! d (Exception. "Failed retrying"))
+             d)
+
+           ;; try again, after the waiting period
+           (and (< tries stop) (= ::local-retry-error val))
+           (let [wait (mt/seconds (math/expt p tries))]
+             (mt/in wait #(md/recur (inc tries))))))))))
```
lvh: this feels like it could be abstracted a little further, e.g. a fn that takes the number of tries and returns a number of seconds to retry or fail/raise an exception)

src/cloudpassage_lib/core.clj
derwolfe: This is local to the function - it shouldn't leak out. That being said, I'd be open to changing this.

src/cloudpassage_lib/core.clj
```diff
@@ -100,7 +147,8 @@
       ;; a token is in redis
       (fernet/decrypt fernet-key token)
       ;; no token is present, fetch a new one
-      (let [new-token (get-auth-token! client-id client-secret)
+      ;; the operation to fetch the token should be asynchronous
```
lvh: What does this comment mean? If it means what I think it means, perhaps this should use let-flow + md/future?

derwolfe: This was a holdover from a todo - removing. But, I do think `fetch-token!` should be asynchronous, it just didn't seem like I should handle it in this PR. What do you think?

lvh: Sure, that's fine -- but then there should be a ticket for that :)

src/cloudpassage_lib/core.clj
derwolfe: @lvh - I think I might need a bit of help to get this working the way you've requested. I've put up a [WIP PR](https://github.com/RackSec/cloudpassage-lib/pull/49/files#diff-b061a2d01bcc7495a5b47994a5e3cba5R82) that contains the current work.

It looks like I'm able to cause the deferred to enter an error state by instantiating an exception inside of any of the functions. But, a consuming function, `get-auth-token` doesn't seem to be handling this as I'd expect - namely, it is exploding with a json-parsing error, instead of exploding with the exception generated at [line 82](https://github.com/RackSec/cloudpassage-lib/pull/49/files#diff-b061a2d01bcc7495a5b47994a5e3cba5R82)

derwolfe: The test [failure](https://travis-ci.org/RackSec/cloudpassage-lib/builds/122519163#L229) I'm talking about is generated by this [test](https://github.com/RackSec/cloudpassage-lib/pull/49/files#diff-05384040988608db6ad94455765af8d9R71).

src/cloudpassage_lib/core.clj
derwolfe: Ticket: https://github.com/RackSec/cloudpassage-lib/issues/50

src/cloudpassage_lib/core.clj
```diff
@@ -44,23 +45,70 @@
...
+               (error "Failed retrying" tries "times; stopping")
+               (throw (Exception. "Failed retrying; stopping.")))
+
+             ;; return the value, don't retry
+             (not errval?) val
+
+             ;; keep retrying
+             (and (< tries stop) errval?)
+             (let [wait (mt/seconds (math/expt p tries))]
+               (mt/in wait #(md/recur (inc tries)))))))))))
```
derwolfe: Comment lost: need to abstract this wait into an `fn`

src/cloudpassage_lib/core.clj
```diff
@@ -44,23 +45,70 @@
...
+                   (md/catch
+                    Exception
+                    (fn [exc]
+                      (error "Failure retrying:" (.getMessage exc))
+                      exc))))]
+    (md/loop [tries 1]
+      (md/chain
+       (invoker)
+       (fn [val]
+         (let [errval? (instance? Exception val)]
```
derwolfe: @lvh: I *think* this is more consistent with what you wanted - what do you think?

lvh: Hm. Why isn't the md/catch moved inside the loop here?

derwolfe: The main reason is that I only want to catch exceptions generated by the invoker - currently the `catch` *is* in the chain, it is just catching exceptions generated by `f`. Were it nested inside of the root `chain` (the outer level), it would catch the exception thrown for `Failed retrying`.

lvh: Is that true if the catch comes before the fn that starts at line 75?

derwolfe: Yes. The only way I know to `catch` an exception with manifold is to execute `f` inside of the catch; that is being doing here using the thread first macro. I've not been able to structure it like the following and have it execute successfully.

```clojure
      (md/chain
       (f)
       (md/catch
           Exception
           (fn [exc]
             (error "Failure retrying:" (.getMessage exc))
             exc))
       (fn [val]
```
And have it execute successfully.

lvh: Huh. That's weird. Just reading how the evaluation works, it looks like the two should be identical...

derwolfe: lvh: just to make sure I'm understanding - your point is the two following functions should execute the same way?

with thread-first
```clojure
(defn retry-tf
  [f p stop]
  (let [recur-in (fn [tries]
                   (let [wait (mt/seconds (math/expt p tries))]
                     (mt/in wait #(md/recur (inc tries)))))]
    (md/loop [tries 1]
      (md/chain
       (->
        (f)
        (md/catch
            Exception
            (fn [exc]
              (error "Failure retrying:" (.getMessage exc))
              exc)))
       (fn [val]
         (let [errval? (instance? Exception val)]
           (cond
             ;; stop trying
             (and (= tries stop) errval?)
             (do
               (error "Failed retrying" tries "times; stopping")
               (throw (Exception. "Failed retrying; stopping.")))

             ;; return the value, don't retry
             (not errval?) val

             ;; keep retrying
             (and (< tries stop) errval?)
             (recur-in tries))))))))
```

without thread-first
```clojure
(defn retry-no-tf
  [f p stop]
  (let [recur-in (fn [tries]
                   (let [wait (mt/seconds (math/expt p tries))]
                     (mt/in wait #(md/recur (inc tries)))))]
    (md/loop [tries 1]
      (md/chain
       (f)
       (md/catch
           Exception
           (fn [exc]
             (error "Failure retrying:" (.getMessage exc))
             exc))
       (fn [val]
         (let [errval? (instance? Exception val)]
           (cond
             ;; stop trying
             (and (= tries stop) errval?)
             (do
               (error "Failed retrying" tries "times; stopping")
               (throw (Exception. "Failed retrying; stopping.")))

             ;; return the value, don't retry
             (not errval?) val

             ;; keep retrying
             (and (< tries stop) errval?)
             (recur-in tries))))))))
```

lvh: No; thread-first will pass the value of `(f)` to md/chain, which the second one doesn't.

derwolfe: How about this?

```clojure
(defn retry
  [f p stop]
  (let [recur-in (fn [tries]
                   (let [wait (mt/seconds (math/expt p tries))]
                     (mt/in wait #(md/recur (inc tries)))))]
    (md/loop [tries 1]
      (md/chain
       (md/catch
           (f)
           Exception
           (fn [exc]
             (error "Failure retrying:" (.getMessage exc))
             exc))
       (fn [val]
         (let [errval? (instance? Exception val)]
           (cond
             ;; stop trying
             (and (= tries stop) errval?)
             (do
               (error "Failed retrying" tries "times; stopping")
               (throw (Exception. "Failed retrying; stopping.")))

             ;; return the value, don't retry
             (not errval?) val

             ;; keep retrying
             (and (< tries stop) errval?)
             (recur-in tries))))))))
```
It removes invoker and pulls catch into the chain?

lvh: Is the only purpose of wrapping the error in that catch so that it turns into a value? If so, why not deal with the retry logic in the catch directly?

(Also "recur-in" feels like it takes a number of seconds; maybe that should have a different name)

derwolfe: Yes! Making those change now - it should be much clearer this way.

derwolfe:
@lvh - this should be more in line with what you suggested regarding catch. :-).

src/cloudpassage_lib/core.clj
```diff
@@ -44,23 +45,57 @@
...
+      (md/catch
+       (f)
+       Exception
+        (fn [exc]
+          (error "Failure retrying:" (.getMessage exc))
+          (if (= tries stop)
+            (do
+              (error "Failed retrying" tries "times; stopping")
+              (throw (Exception. "Failed retrying; stopping.")))
+            (try-again tries)))))))
```
lvh: Some general thoughts based on trying to build this for otter which you are free to ignore:

- Not every exception warrants a retry; retry logic needs to know the exception type to decide (sometimes). Also, not every exception warrants retrying as often.
- Providing the actual exception seems significantly more useful than the message of the exception. Knowing that it was retried is additional information that's generally implicitly known to the caller anyway; the underlying exception is what you really care about.
- If "try-again" is to be extracted; that's probably not the fn you want to know about `mt/in` or `md/recur` -- that should probably stay an internal detail of this fn.
- You can compose behavior with combinators:

```clojure
(defn up-to
  [max-tries retry?]
  (fn [tries e]
    (when (< tries max-tries)
      (retry? tries e))))

(defn exponentially
  [period]
  (fn [tries _]
    (math/expt period tries)))

(defn circuit-breaker
  [retry? cls]
  (fn [tries e]
    (when-not (instance? cls e)
      (retry tries e))))

(->> (exponentially 2)
     (circuit-breaker SuperBadTLSProblem)
     (up-to 10)) ;; 2 ** 10 => 1024s
```

Otter at one point had a nasty problem where it needed to know if an exception `b` occurred after an exception `a` occurred;  because that was a special case. That would mean keeping track of tries so far and passing that to the above fn. That's a pretty exotic use case though :)

lvh: (although if you keep a list of excs around, finding the current try count is just `(count failures)` away :))

derwolfe: Thanks for your thougtful comments, @lvh

> - Not every exception warrants a retry; retry logic needs to know
>   the exception type to decide (sometimes). Also, not every
>   exception warrants retrying as often.

I agree with this in general - the only reason I'm catching everything
is that I've only seen `aleph` throw instances of `Exception` that
then provide a more specific error message - e.g., Exception is thrown
with a `401` or a `500` as part of the message. It doesn't seem like
it has provided more specific exception types which we could catch.

> - Providing the actual exception seems significantly more useful than
>  the message of the exception. Knowing that it was retried is
>  additional information that's generally implicitly known to the
>  caller anyway; the underlying exception is what you really care
>  about.

Thanks, I'll change this.
    
> - If "try-again" is to be extracted; that's probably not the fn you
>   want to know about `mt/in` or `md/recur` -- that should probably
>   stay an internal detail of this fn.

I agree that extracting both `md/in` and `md/recur` makes this less
clear; for now I'll move back to keeping everything in the function
body. But, if they stay in the main part of the function, I'm not sure
what else could be extracted from the recurring step, unless this
moves into using the combinators you've suggested)

> (although if you keep a list of excs around, finding the current try
> count is just (count failures) away :))

Would you recommend logging each of the exceptions one by one or providing a
dump of the exceptions list on the final failure?

lvh: Re: aleph returning Exception only, yes (I don't actually know, I'm assuming you're correct), but:

- If we asynchronify a different library, it will return arbitrary exceptions
- Sorry, exception type was a red herring; "underlying cause" would have been a better word. The function I'm suggesting gets the exception instance, so it can still dispatch on details, e.g. 404 (no point retrying, for example) vs 500 (just try again)

Re: what can we extract? => that fn should take a number of tries (and maybe exception instance) and return nil (stop trying) or a number of seconds within which to try again... But then you probably just want the combinators.

Re: combinator API: I think we should leverage the list monad to encapsulate the uncertainty in the number of errors.

HHOS; don't worry I did not turn into a Haskell programmer; OTOH the exception list API is a little weird but it does give you everything you need; most recent exception is just `(last es)`, number of tries `(count es)`...

I think exceptions should probably be logged as they happen, but communicated on the way out. That's a little trickier than it looks at first: Throwables can't generally have metadata attached to them. Java has a `cause` thing for Throwables that's intended for this, but it takes exactly 1 Throwable, whereas you probably want to communicate the list.

derwolfe:
It might make sense to break this up into a few prs:

1) Bump manifold
2) Modify `get-auth-token!` so that it returns a deferred, instead of blocking on dereferencing the value.
3) Add retry logic/tests
4) Use retry logic inside with `get-auth-token!`

PR #49, opened 2016-04-12
[NO MERGE] Deferred err handling

derwolfe:
This is for discussion only.

### DIFF
--- a/src/cloudpassage_lib/core.clj
+++ b/src/cloudpassage_lib/core.clj
@@ -46,8 +46,8 @@
   (f/unparse cp-date-formatter date))
 
 (defn retry
-  "Takes a function that returns a deferred. Retries that function
-  until it succeeds or the number of failures equal the stop value
+  "Takes a function that returns a `manifold.deferred/deferred`. Retries that
+  function until it succeeds or the number of failures equal the stop value.
 
   `retry` expects to encounter exceptions when retrying `f`. As such,
   it will catch all exceptions that `f` might throw and continue retrying.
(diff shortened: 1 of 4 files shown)


src/cloudpassage_lib/core.clj
```diff
@@ -44,23 +45,71 @@
...
+      (md/chain
+       (invoker)
+       (fn [val]
+         (let [errval? (instance? Exception val)]
+           (cond
+             ;; stop trying
+             (and (= tries stop) errval?)
+             (do
+               (error "Failed retrying" tries "times; stopping")
+               (Exception. "Failed retrying; stopping."))
```
lvh: Hm. I thought you were supposed to raise, and that the deferred machinery catches it; but I may be confusing with Twisted :)

lvh: e.g. for chain, this is true: you either return a deferred that is in an error state, or you raise.

lvh: Oh, hm. Reading the documentation, that looks like it might be chain-specific behavior. Worth a shot, though.

derwolfe: What I'm finding a bit confusing is that if I use raise at 82, it doesn't seem like manifold captures it - it is like it just does a bare raise that clojure handles.

lvh: That's definitely a bug then; see section starting with "If any stage in chain throws an exception"... at https://github.com/ztellman/manifold/blob/master/docs/deferred.md

derwolfe: hmm - That could be an artifact of me derefing in the test. That would unwrap the exception and cause it show up. I'll experiment a bit further.

derwolfe: That was the issue :-). My bug.

PR #51, opened 2016-04-14
Return deferred from `get-auth-token!`

derwolfe:
Modify `get-auth-token!` to return a deferred wrapping a token, instead of the actual token.

This is a PR broken out of #48.

### DIFF
--- a/test/cloudpassage_lib/redis_test.clj
+++ b/test/cloudpassage_lib/redis_test.clj
@@ -5,13 +5,18 @@
             [taoensso.carmine :as car]
             [fernet.core :as fernet-clj]
             [cloudpassage-lib.fernet :as cp-fernet]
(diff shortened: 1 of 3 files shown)


test/cloudpassage_lib/redis_test.clj
```diff
@@ -39,9 +44,7 @@
                         :access_token "12"}
             expected-token (:access_token fake-token)
             encrypted-token (cp-fernet/encrypt fernet-key expected-token)
-            get-fake-token (fn [_ _]
-                             (reset! called false)
-                             "if called, I should explode")]
+            get-fake-token (fn [_ _] (reset! called true))]
```
ehashman: Note: this test was broken before.

ehashman:
LGTM!

PR #53, opened 2016-04-14
Add exponential backoff logic and tests

derwolfe:
Broken out of #48.

This adds the implementation of the exponential backoff/retry logic.

### DIFF
--- a/src/cloudpassage_lib/retry.clj
+++ b/src/cloudpassage_lib/retry.clj
@@ -1,0 +1,57 @@
+(ns cloudpassage-lib.retry
+  (:require
+   [clojure.math.numeric-tower :as math]
+   [manifold.deferred :as md]
+   [manifold.time :as mt]))
+
+(defn exponentially
+  "Returns a function that when evaluated will produce the initial wait
+  raised to the number of failures.
+
+  This is intended to be used in conjunction with other 1-arg combinators,
+  that take a vector of failures."
+  [wait]
+  (fn [failures]
+    (math/expt wait (count failures))))
+
+(defn up-to
+  "Returns a function that when evaluated will either
+
+  1) return a number to use to wait until retrying a function again. Or,
+  2) throw an exception because the maximum number of retries, `stop`, has
+  been reached.
+
+  This is intended to be used in conjunction with other 1-arg combinators,
+  that take a vector of failures."
+  [stop retry?]
+  (fn [failures]
+    (if (< (count failures) stop)
+      (retry? failures)
+      (throw (last failures)))))
+
+(defn retry-exp-backoff
+  "Takes a function that returns a `manifold.deferred/deferred`. Retries that
+  function until it succeeds or the number of failures equal the stop value.
+
+  Expects to encounter exceptions when retrying `f`. As such,
+  it will catch all exceptions that `f` might throw and continue retrying.
+
+  f - a function that should be retried; must return a
+      `manifold.deferred/deferred'
+  p - an int representing the initial number of seconds to wait before retrying.
+      This will grow exponentially for each attempt.
+  stop - an int representing the number of tries that the api should make before
+         giving up and returning the last exception encountered.
+
+  Returns a deferred wrapping the results of `f`."
+  [f p stop]
+  (let [try-after (->> (exponentially p)
+                       (up-to stop))]
+    (md/loop [failures []]
+      (md/catch
+       (f)
+       Exception
+        (fn [exc]
+          (let [all-failures (conj failures exc)
+                wait (try-after all-failures)]
+            (mt/in (mt/seconds wait) #(md/recur all-failures))))))))
--- a/src/cloudpassage_lib/retry.clj
+++ b/src/cloudpassage_lib/retry.clj
@@ -8,8 +8,8 @@
   "Returns a function that when evaluated will produce the initial wait
   raised to the number of failures.
 
-  This is intended to be used in conjunction with other 1-arg combinators,
-  that take a vector of failures."
+  This is intended to be used in conjunction with 1-arg combinators that take a
+  vector of failures."
   [wait]
   (fn [failures]
     (math/expt wait (count failures))))
@@ -21,8 +21,8 @@
   2) throw an exception because the maximum number of retries, `stop`, has
   been reached.
 
-  This is intended to be used in conjunction with other 1-arg combinators,
-  that take a vector of failures."
+  This is intended to be used in conjunction with 1-arg combinators that take a
+  vector of failures."
   [stop retry?]
   (fn [failures]
     (if (< (count failures) stop)
--- a/src/cloudpassage_lib/retry.clj
+++ b/src/cloudpassage_lib/retry.clj
@@ -4,7 +4,7 @@
    [manifold.deferred :as md]
    [manifold.time :as mt]))
 
-(defn exponentially
+(defn ^:private exponentially
   "Returns a function that when evaluated will produce the initial wait
   raised to the number of failures.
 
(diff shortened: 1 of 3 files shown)


derwolfe:
Codecov should be passing, it just doesn't know how to handle the `manifold.deferred/loop` macro.

src/cloudpassage_lib/retry.clj
```diff
@@ -0,0 +1,57 @@
...
+  (let [try-after (->> (exponentially p)
+                       (up-to stop))]
+    (md/loop [failures []]
+      (md/catch
+       (f)
+       Exception
+        (fn [exc]
+          (let [all-failures (conj failures exc)
+                wait (try-after all-failures)]
+            (mt/in (mt/seconds wait) #(md/recur all-failures))))))))
```
lvh: This all looks good to me; my only preference would be that retry-exp-backoff is implemented as `(partial retry (->> (exponentially p) (up-to stop)))`, and that's a nitpick that you can handle separately if you'd like. It just feels like this did most of the hard work to get it decomplected, and then not get the prize at the end :)

derwolfe: I feel like I'm getting closer to being able to provide that as the implementation; but I'm missing something. This is the closest I've been able to get:

```clojure
(defn retry
  [strategy f]
  (md/loop [failures []]
    (md/catch
        (f)
        Exception
      (fn [exc]
        (let [all-failures (conj failures exc)
              wait (strategy all-failures)]
          (mt/in (mt/seconds wait) #(md/recur all-failures)))))))

(defn retry-exp-backoff
  "Takes a function that returns a `manifold.deferred/deferred`. Retries that
  function until it succeeds or the number of failures equal the stop value.

  Expects to encounter exceptions when retrying `f`. As such,
  it will catch all exceptions that `f` might throw and continue retrying.

  f - a function that should be retried; must return a
      `manifold.deferred/deferred'
  p - an int representing the initial number of seconds to wait before retrying.
      This will grow exponentially for each attempt.
  stop - an int representing the number of tries that the api should make before
         giving up and returning the last exception encountered.

  Returns a deferred wrapping the results of `f`."
  [f p stop]
  (retry (->> (exponentially p)
              (up-to stop))
         f))
```

I've experimented with trying to have `retry-exp-backoff` return a partial, where `retry-exp-backoff` is still a function; but then callers of `retry-exp-backoff` don't receive the result of the retry, they are given the partial function. 

This confuses me a bit - in order to use partial, it seems like `retry-exp-back` should be implemented as a function, instead of just a being a binding to a partial - otherwise, how can the parameters `p` and `stop` be bound to the partial, but `f` be passed in later?

lvh: Partial builds the fn for you, so you use def, not defn. That said, this implementation is totally fine.

derwolfe: I like this idea as well, but the closest I seem to be able to get is:

```clojure
(defn retry
  [strategy f]
  (md/loop [failures []]
    (md/catch
        (f)
        Exception
      (fn [exc]
        (let [all-failures (conj failures exc)
              wait (strategy all-failures)]
          (mt/in (mt/seconds wait) #(md/recur all-failures)))))))

(defn retry-exp-backoff
  "Takes a function that returns a `manifold.deferred/deferred`. Retries that
  function until it succeeds or the number of failures equal the stop value.

  Expects to encounter exceptions when retrying `f`. As such,
  it will catch all exceptions that `f` might throw and continue retrying.

  f - a function that should be retried; must return a
      `manifold.deferred/deferred'
  p - an int representing the initial number of seconds to wait before retrying.
      This will grow exponentially for each attempt.
  stop - an int representing the number of tries that the api should make before
         giving up and returning the last exception encountered.

  Returns a deferred wrapping the results of `f`."
  [f p stop]
  (retry (->> (exponentially p)
              (up-to stop))
         f))
```

lvh: Yeah, I totally brain farted there. Sorry! that Implementation is fine.

lvh: (The partial would only work if the retry logic was fixed)

derwolfe: No worries - I'll commit the implementation with retry being extracted and change the tests. Also, I might privatize everything except for `retry-exp-backoff`...

src/cloudpassage_lib/retry.clj
```diff
@@ -0,0 +1,70 @@
...
+  (:require
+   [clojure.math.numeric-tower :as math]
+   [manifold.deferred :as md]
+   [manifold.time :as mt]))
+
+(defn ^:private exponentially
+  "Returns a function that when evaluated will produce the initial wait
+  raised to the number of failures.
+
+  This is intended to be used in conjunction with other 1-arg combinators,
```
lvh: Minor nitpick: this appears to suggest that exponentially itself is a combinator, which isn't true; so `s/other //`?

src/cloudpassage_lib/retry.clj
```diff
@@ -0,0 +1,70 @@
...
+
+(defn ^:private up-to
+  "Returns a function that when evaluated will either
+
+  1) return a number to use to wait until retrying a function again. Or,
+  2) throw an exception because the maximum number of retries, `stop`, has
+  been reached.
+
+  This is intended to be used in conjunction with other 1-arg combinators,
+  that take a vector of failures."
```
lvh: Similar nitpick: combinators are tools that produce things that take vectors of failures; they do not take the vector of failures themselves?

derwolfe:
@lvh: Thanks for the comments - they've been addressed :-)

test/cloudpassage_lib/retry_test.clj
```diff
@@ -0,0 +1,89 @@
...
+      (is (= 1000 (f [:a :b :c]))))))
+
+(deftest up-to-tests
+  (testing "raises most recent exception when number of tries exceeded"
+    (let [tries [(Exception. "earlier")
+                 (Exception. "recent")]
+          stop 2
+          retry? #(throw (Exception. "I shouldn't have been called"))
+          f (#'retry/up-to stop retry?)]
+      (is (thrown-with-msg? Exception #"recent" (f tries)))))
```
lvh: I wish there was a way to communicate to cloverage that you are expecting retry? not to be called.

lvh:
I really like this API & implementation. I can't wait to use it all over the place. Great job :D

derwolfe:
Thanks a lot for your help! I like it too :-).

src/cloudpassage_lib/retry.clj
```diff
@@ -0,0 +1,70 @@
...
+      exceptions. The number of exceptions is used to record the number
+      of attempts.
+
+  Returns a deferred wrapping the results of `f`."
+  [f strategy]
+  (md/loop [failures []]
+    (md/catch
+     (f)
+     Exception
+      (fn [exc]
```
ehashman: Why is this function indented by one additional space (and why didn't cljfmt complain about it?)

derwolfe: I don't know, but cljfmt *corrected* the function to look this way.

lvh: hint: try indenting a try/catch

derwolfe: It seems odd to me too though - my guess is that it doesn't handle the namespaced `md/loop` and `md/catch` sections correctly.

src/cloudpassage_lib/retry.clj
```diff
@@ -0,0 +1,70 @@
...
+      `manifold.deferred/deferred'
+  p - an int representing the initial number of seconds to wait before retrying.
+      This will grow exponentially for each attempt.
+  stop - an int representing the number of tries that the api should make before
+         giving up and returning the last exception encountered.
+
+  Returns a deferred wrapping the results of `f`."
+  [f p stop]
+  (retry f (->> (exponentially p)
+                (up-to stop))))
```
ehashman: This is a little bit confusing for me to follow with all the currying but it looks really pretty \o/

PR #54, opened 2016-04-19
retry authing using the new retry logic

derwolfe:
Use the new retry logic when authenticating with apis.

Minor cleanup:
1) Instead of using a function that returns a success/error deferred using `md/success!` or `md/error!`, use the utility functions `error-deferred` or `success-deferred` that build a return in either state and return it.
2) Doc fixes around the new api.


lvh:
LGTM!

PR #59, opened 2016-05-10
Use new fernet

derwolfe:
`fernet-clj` now exposes `encrypt-string` and `decrypt-to-string` so this library no longer needs implement those functions.


project.clj
```diff
@@ -11,7 +11,7 @@
                  [clj-time "0.11.0"]
                  [base64-clj "0.1.1"]
                  [com.taoensso/carmine "2.12.1"]
-                 [fernet "0.1.0"]
+                 [fernet "0.2.0" :exclusions [io.netty/netty-buffer]]
```
ehashman: May I ask why you're excluding this? (presumably because it's not used anywhere and so we don't want it in the uberjar, but wanted to confirm)

ehashman:
This lgtm but I want to test it against clark-kent before merging, one moment...

project.clj
derwolfe: Sure :-), there is a collision between the version that fernet-clj is using and what aleph is using.

ehashman:
REPL warnings

```
Warning: protocol #'clojurewerkz.buffy.types.protocols/BuffyType is overwriting function read
WARNING: read already refers to: #'clojure.core/read in namespace: clojurewerkz.buffy.types.protocols, being replaced by: #'clojurewerkz.buffy.types.protocols/read
WARNING: read already refers to: #'clojure.core/read in namespace: fernet.frame, being replaced by: #'clojurewerkz.buffy.types.protocols/read
```

Otherwise, looking good.

derwolfe:
I'll close this for now until I fix the warnings generated upstream.

PR #63, opened 2016-05-19
Fix get-page! error handling, use exponential backoff on scans fetches

### LINKED ISSUE #61: Fix error-handling anywhere `get-page!` is used
`scans/get-page!` throws an Exception, but if an exception is thrown inside a manifold deferred, manifold catches this exception and prints out a stack trace instead. I suspect what we actually want to do is to `md/catch` the Exception before manifold does and throw another one ourselves, in order for that to propagate to the report consumer.

ehashman:
Fixes #61.

### DIFF
--- a/src/cloudpassage_lib/scans.clj
+++ b/src/cloudpassage_lib/scans.clj
@@ -48,9 +48,10 @@
         #(md/chain
           (cpc/get-single-events-page! token url)
           (fn [response]
-            (if (cpc/page-response-ok? response) response
-                (do (warn "Couldn't fetch page.")
-                    (throw (Exception. "No more retries."))))))]
+            (if (cpc/page-response-ok? response)
+              response
+              (do (warn "Couldn't fetch page.")
+                  (throw (Exception. "No more retries."))))))]
     (retry/retry-exp-backoff
      get-events-page-or-throw
      timeout
--- a/src/cloudpassage_lib/scans.clj
+++ b/src/cloudpassage_lib/scans.clj
@@ -11,7 +11,8 @@
    [taoensso.timbre :as timbre :refer [warn error info]]
    [camel-snake-kebab.core :as cskc]
    [camel-snake-kebab.extras :as cske]
-   [cloudpassage-lib.core :as cpc]))
+   [cloudpassage-lib.core :as cpc]
+   [cloudpassage-lib.retry :as retry]))
 
 (def ^:private base-scans-url
   "https://api.cloudpassage.com/v1/scans/")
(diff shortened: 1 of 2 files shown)


src/cloudpassage_lib/scans.clj
```diff
@@ -42,25 +43,36 @@
...
-       (do (warn "Couldn't fetch page. Retrying.")
-           (mt/in
-            timeout
-            #(get-page-retry! token url (dec num-retries) timeout)))))))
+  (let [num-tries (inc num-retries)
+        get-events-page-or-throw
+        #(md/chain
+          (cpc/get-single-events-page! token url)
+          (fn [response]
+            (if (cpc/page-response-ok? response) response
```
lvh: I misread this at first, but I'm guessing the linter complained about it already :)

ehashman: The linter didn't complain about this one; it complained about one of the tests where a value is generated but not used.

How could I make this more legible?

lvh: `response` should be on its own line, to clarify that there are actually 2 branches to this `if`

src/cloudpassage_lib/scans.clj
```diff
@@ -42,25 +43,37 @@
...
+              response
+              (do (warn "Couldn't fetch page.")
+                  (throw (Exception. "No more retries."))))))]
+    (retry/retry-exp-backoff
+     get-events-page-or-throw
+     timeout
+     num-tries)))
+
+(defn ^:private handle-stream-exception
+  [deferred output-stream input-stream]
```
lvh: I'm not sure why GitHub is rendering this as a highlighted keyword.

test/cloudpassage_lib/scans_test.clj
```diff
@@ -37,24 +37,35 @@
...
-          (is (str/includes? (last @log) "No more retries."))))))
+          (doseq [logline @log]
+            (is (str/includes? logline "Couldn't fetch page.")))))
+      (testing "Exception thrown in report isn't handled by manifold"
+        (with-redefs [cpc/fetch-token! (constantly "yay")]
+          (let [log (use-atom-log-appender!)]
+            (is (thrown-with-msg?
+                 Exception
+                 #"Report failed to generate"
+                 (vec (scans/fim-report! '_ '_)))) ;; make lazyseq eval
```
lvh: The linter complains because the return value is discarded. `dorun`, `doall`, `doseq` are the ways to go through a seq for side effects; in this case, since you don't care about the result, `dorun` appears to be the best solution.

lvh:
LGTM modulo comment (which the linter looks like it can catch)

ehashman:
CI is now happy :)

PR #66, opened 2016-05-24
Remove 'error' from warning, avoiding false alerts

ehashman:
This is just a quick fix to help with alerting/monitoring. Currently warnings are generating alerts because they contain the text "error", so this fixes that.


fboxwala:
lgtm

PR #68, opened 2016-06-07
Die if redis environment variables aren't defined

ehashman:
Partially addresses #12.

After seeing @derwolfe get confused by this on yet another PR, I decided it was time to fix this. When REDIS_URL and REDIS_TIMEOUT environment variables aren't set, cloudpassage-lib will log a fatal error and exit, as it cannot operate without these defined.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -16,6 +16,7 @@
                  [cheshire "5.5.0"]
                  [camel-snake-kebab "0.3.2"]
                  [org.clojure/math.numeric-tower "0.0.4"]
+                 [com.gfredericks/system-slash-exit "0.2.0"]
                  [com.cemerick/url "0.1.1"]]
   :plugins [[lein-auto "0.1.2"]
             [lein-cljfmt "0.3.0"]
--- a/src/cloudpassage_lib/core.clj
+++ b/src/cloudpassage_lib/core.clj
@@ -6,7 +6,8 @@
    [manifold.deferred :as md]
    [byte-streams :as bs]
    [taoensso.carmine :as car :refer (wcar)]
-   [taoensso.timbre :as timbre :refer [info warn]]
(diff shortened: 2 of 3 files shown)


derwolfe:
@ehashman not sure I follow, I was confused about the `clientid:clientsecret:name` for accounts, not redis :)

ehashman:
Ah. I can add a similar thing there. This serves as a nice PoC.

derwolfe:
The issue was that my 'profiles.clj' file was out of date and didn't contain a 'name' in the 'accounts' string. This caused an NPE in Clark-Kent because it relied on the name being present.

derwolfe:
@ehashman - I like this idea :). My suggestion for the applications (reporter, etc.) would be to create a function that validates all of the necessary configuration variables, logs the errors for each undefined var, and exits. This would be something we could call in main/whatever our startup function is and could be tested relatively easily. 

Something similar could be done with this library, although I'm not sure when it should run (at access time, at the top, etc?). It looks like the values needed to run cloudpassage-lib are a `fernet-key`, `redis-url`, and `redis-timeout`. I'd suggest we consolidate the checks for all the vars in a single function and perform the logging and call to exit there.

ehashman:
Ahhh, I missed the `fernet-key`. I'll have to take a look at the code to see where the best place to put this validation is. Since this is a library, there isn't really a main here where we could do validation. But I also feel like it should be the library's responsibility to verify any environment variables it needs internally (as opposed to at the consumer level/main) because these could always change without the consumer knowing. For instance, I think we had some plans of factoring that out entirely such that the consumer had to pass them in?

ehashman:
I'm going to close this until I can figure out a better way to do this. Since it sounds like we need to do the full refactor.

PR #69, opened 2016-06-09
remove cloudpassage.retry, use banach

derwolfe:
The retry logic has been factored out and placed in the open source [banach](https://github.com/racksec/banach) project.


ehashman:
Since we now have a copy of test_utils.clj under the banach project, do we also want to remove that from this project to improve code reuse?

ehashman:
(we have a copy of `use-atom-log-appender!` in four different projects now, including banach, so it would be nice to reduce that number to ideally 1)

derwolfe:
@ehashman I agree, it would be nice if it were in a central location. That being said, the log appender test helper isn't something I was thinking that banach would be guaranteeing in terms of public api.

ehashman:
I am also okay with the number being e.g. 2, such that the projects that are dependent on cloudpassage-lib can just use its version, and it's fine if banach rolls its own. But 4 seems silly. :P

PR #72, opened 2016-12-07
Upgrade dependencies, prep release

ehashman:


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.2.7
+
+- Update dependencies
+- Hopefully the new `aleph` will address some of the intermittent livelocking
+  we've seen in production
```
derwolfe: What livelocking? Has this been reporting upstream?

ehashman: @derwolfe: https://rackspace.slack.com/archives/rms-eng-internal/p1473781030000031

TL;DR: occasionally, aleph + netty would get into a livelock state and reports would not complete. I couldn't figure out the conditions to reproduce (it happened about once a month across 33 reports, so close to a 1/1000 likelihood), hence I was unable to write up an actionable report for upstream.

I talked to some people at Clojure/conj who had also encountered a similar issue around intermittent livelocking with aleph that was difficult to reproduce but may have been related to low resource utilization (which sounded familiar). They suggested upgrading.

derwolfe: thanks!

PR #73, opened 2017-01-09
[WIP] Drop redis dependency

derwolfe:
This removes redis as a dependency on the library. Previously, API tokens were cached in redis. This functionality has been replaced by the use of a `clojure.core.cache/TTLCache` wrapped in an atom. This does mean that the library still has an opinion with respect to the caching of tokens, _but_ this is opt-in, if consumers of the library would prefer to handle the token fetching/caching logic themselves, they need only use `cloudpassage-lib.core/get-auth-token!`. 

At some later date, we might look at dropping support for `fetch-token!` entirely.


README.md
```diff
@@ -9,52 +9,12 @@ A Clojure library for interacting with CloudPassage APIs.
...
-  ```clojure
-  {:dev {:env {:redis-url "redis://localhost:6379"
-               :redis-timeout "4000"
-               :fernet-key "AVALIDFERNETKEY"}}}
-  {:test {:env {:redis-url "redis://localhost:6379"
-                :redis-timeout "4000"}}}
-  ```
-
-Now you are able to make API calls such as `cloudpassage-lib.scans/fim-report!`.
-
```
ehashman: AMAZING

derwolfe:
So _anything_ depending on `fetch-token` _will_ need to change. The signature has been reduced from `client-id, client-secret, fernet-key` to `client-id, client-secret`.

PR #76, opened 2017-02-07
Quiet down the logs

ehashman:
Just tested the new version of cloudpassage-lib with clark-kent, and every single time the cache is checked a log message is printed. This makes the logs a little noisier than I'd like.

I've moved these messages to TRACE level, which aren't printed by default.


ehashman:
Wow codecov is rood

derwolfe:
Codecov also seems wrong. Merging.

PR #78, opened 2017-06-12
Increase cache TTL from 8sec to 8min

### LINKED ISSUE #77: Increase cache TTL for auth tokens
I was wondering why we need to fetch a new auth token every 4-5 requests in clark-kent, and realized this is because the cache TTL is set to 8000ms (i.e. 8 seconds): https://github.com/RackSec/cloudpassage-lib/blob/48eb2c6ee7840665a8a63cbf6719527a3fcb4ab4/src/cloudpassage_lib/core.clj#L83

The Halo docs suggest the tokens usually live for 15 minutes, although we can check the exact token lifetime in the response body. Is it possible to cache these for 8 minutes (which I believe we did in redis) instead?

ehashman:
Closes #77.

### DIFF
--- a/src/cloudpassage_lib/core.clj
+++ b/src/cloudpassage_lib/core.clj
@@ -80,7 +80,7 @@
   [response]
   (not= ::fetch-error response))
 
-(def ^:private cache-ttl-milliseconds 8000)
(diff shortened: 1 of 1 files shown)


ehashman:
Released.

PR #79, opened 2017-09-29
Attach server to scans.

sirsean:
The data that comes back with scans has some info about the server,
including its hostname. But there is additional data about the server
that we may want, including the EC2 instance-id (if it exists). We
already have the server data, so this adds zero requests. It just uses
slightly more RAM now because that server data is attached to each scan.


derwolfe:
LGTM and seems totally reasonable and is only additive (that is, the new api shouldnt break any consumers) :+1:

